PR #1, opened 2019-01-31
Server config file permission checking

1000TurquoisePogs:
Checking server config file to refuse to start if group has write/execute and other has any


1000TurquoisePogs:
Now depends on https://github.com/zowe/zowe-common-c/pull/5

PR #7, opened 2019-02-06
Warn if not localhost

1000TurquoisePogs:
Since our server does not implement TLS, if the user tries to run the server such that it is going to listen on an externally accessible IP, the server should print a warning alerting them that this better not be for production...


1000TurquoisePogs:
Behavior:
```
Host name is DNS or non-numeric, c0a8371c
*** WARNING: Server doesn't implement HTTPS! ***
*** In production only use localhost or 127.0.0.1! Server using: rs28 ***
ZSS server settings: address=rs28, port=11224
privilegedServerName is not specified, or is NULL.
```

Issue #11, opened 2019-02-22
Add http no-cache headers to i/o apis


1000TurquoisePogs:
/unixfile, /datasetcontents, /datasetmetadata and /vsamdatasetcontents serve potentially sensitive information, and therefore shouldn't be cached by the browser.
To fix, we'd need to add the headers "Cache-control: no-store" and "Pragma: no-cache" to these services in order to secure them.

PR #12, opened 2019-02-22
Add headers for no http caching

1000TurquoisePogs:
Implements https://github.com/zowe/zss/issues/11
Aside from adding headers instructing receivers not to cache the results, this change also unifies the content type of json. The majority of our code used MIME "text/json", but the official MIME type is "application/json", so hopefully this improves compatibility in edge cases.

### DIFF
--- a/c/authService.c
+++ b/c/authService.c
@@ -114,7 +114,7 @@ static void respond(HttpResponse *res, int rc, const ZISAuthServiceStatus
   jsonPrinter* p = respondWithJsonPrinter(res);
 
   setResponseStatus(res, HTTP_STATUS_OK, "OK");
-  setDefaultJSONRESTHeaders(response);
+  setDefaultJSONRESTHeaders(res);
   writeHeader(res);
   if (rc == RC_ZIS_SRVC_OK) {
     jsonStart(p); {
--- a/c/authService.c
+++ b/c/authService.c
@@ -31,6 +31,7 @@
 #include "bpxnet.h"
 #include "socketmgmt.h"
 #include "zis/client.h"
+#include "httpserver.h"
 
 /*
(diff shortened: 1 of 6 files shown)


c/authService.c
```diff
@@ -114,8 +114,7 @@ static void respond(HttpResponse *res, int rc, const ZISAuthServiceStatus
   jsonPrinter* p = respondWithJsonPrinter(res);
 
   setResponseStatus(res, HTTP_STATUS_OK, "OK");
-  setContentType(res, "application/json");
-  addStringHeader(res, "Transfer-Encoding", "chunked");
+  setDefaultJSONRESTHeaders(res);
```
ifakhrutdinov: authService.c doesn't include dataservice.h. To the compiler the ```setDefaultJSONRESTHeader``` looks like ```int foo(int v)```. If compiled with LP64, the response variable and the return value will always be truncated to 4 bytes, i.e. an int.

These are the only file that have the correct function declaration:
```
./dataservice.u:dataservice.o: /u/foobar/zowe-common-c/h/dataservice.h
./zss.u:zss.o: /u/foobar/zowe-common-c/h/dataservice.h
./authService.u:authService.o: /u/foobar/zowe-common-c/h/dataservice.h
./omvsService.u:omvsService.o: /u/foobar/zowe-common-c/h/dataservice.h
./datasetService.u:datasetService.o: /u/foobar/zowe-common-c/h/dataservice.h
./zosDiscovery.u:zosDiscovery.o: /u/foobar/zowe-common-c/h/dataservice.h
```

Same applies to httpserver.c, which shouldn't include dataservice.h I think but now have to. And anyone who wants to use it will have to link with dataservice.c.

PR #17, opened 2019-03-26
RBAC: Remove the username and class from the url  

fkovinAtRocket:
- /user/class/profile/level -> /profile/level
- use the authenticated user name from the request
- use the hardcoded class ZOWE

I've not yet replaced the call to ZIS with a direct RACROUTE call as this is a lower priority. 

Also, this PR contains an unrelated change that fixes the nasty problem with config file permissions in a dev sandbox.

### DIFF
--- a/c/zss.c
+++ b/c/zss.c
@@ -795,6 +795,17 @@ static int validateConfigPermissionsInner(const char *path) {
   return 0;
 }
 
+#ifdef ZSS_IGNORE_PERMISSION_PROBLEMS
+
+static int validateFilePermissions(const char *filePath) {
+  zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_SEVERE,
+    "Skipping validation of file permissions: disabled during compilation, "
+    "file %s.\n", filePath);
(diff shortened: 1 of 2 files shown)


c/zss.c
```diff
@@ -795,6 +795,17 @@ static int validateConfigPermissionsInner(const char *path) {
   return 0;
 }
 
+#ifdef ZSS_IGNORE_PERMISSION_PROBLEMS
+
+static int validateFilePermissions(const char *filePath) {
```
1000TurquoisePogs: The rest of the changes are OK, but can you please explain this? When/why is this needed? You just need permissions max 750 on the config dir and max 640 on the files in the config dir when running for the original check to pass. That's ../deploy/instance/ZLUX/serverConfig for the standard setup.

fkovinAtRocket: Yeah, exactly, that was very annoying during development that I have to do that sort of thing with my sandbox. I often have to test in nonstandard environments (e.g. upload a minimal set of file to a system like SS03) and that all adds a level of headache. I'm going to do this anyway even if this is never committed.

PR #30, opened 2019-05-02
Side deck for plugins

fkovinAtRocket:
- side deck that a plugin can use to link with the functions in the server
- script that I used to generate it
- header file that has all the functions that are in the side deck


fkovinAtRocket:
Jordan, the file is supposed to be 80 char fixed block, so it doesn't have line endings. Every line is exactly 80 characters, padded with spaces if needed.

fkovinAtRocket:
What really concerns me though is that we're adding many functions whose name don't have prefixes, like: `writeString()`, `writeHeader()`, `readByte()`, `indexOf()`, etc. They might clash with other functions in the plugin code.

Some of them might be non-essential, but the functions in httpserver.c definitely must go there, and they all are unprefixes, with names like `finishResponse()`, `writeBytes()` `writeRequest()` etc. Note that it's absolutely required for a plugin to be able to send HTTP responses.

So I'm wondering if we should take the time and rename them to e.g. `httpServerFinishResponse()`, etc

ifakhrutdinov:
Would we want to include the recovery.c functions (recoveryPush, recoveryPop, etc)?

1000TurquoisePogs:
How far do we go here? charsets.c and xml.c also look like someone may want to use them.

fkovinAtRocket:
1. Please define "useful function". This was my first question and it turned out that nobody knew what exactly it was. It looks like alloc.c might be safely removed, and we don't need all of the stuff in utils.c, but I would argue that logging, recovery and http/dataservice stuff need to be included in full, since they are functions that work with data/state *shared* by ZSS and plugins.
2. I'm not sure how collision-prone names are "automated pitfalls". They're pretty manual pitfalls instead: C doesn't have any namespaces, and any global function/variable name should be reasonably unique. If we're forcing every plugin to link with a symbol called "finishResponse" then chances are moderately high that they might also have a function with that name but a different meaning/arguments/return value: it's just a pretty common name for a function. If it were called "httpServerFinishResponse()" or "zoweHttpFinishResponse()", then, OTOH, chances would be much lower. 

And this particular function meets the definition of a "useful function" at least empirically: it can be observed that most plugins, and, generally HTTP endpoints, use it, to, well, finish a response. And it's also a function that works with data shared by ZSS and plugins, so it has to be part of the side file.

1000TurquoisePogs:
Side files are human-readable files when written to only include functions that plugins should be able to use, and exclude those they shouldn't rely on. They then serve as a form of documentation. What's a "useful" function will change over time, so we should start with a short list and only extend it when requested and the request has merit.

1000TurquoisePogs:
Example of something that shouldn't be in the side file but is in this PR: dequeueHttpRequest

fkovinAtRocket:
It turns out that we actually need xml.c because there's a global function in httpserver that creates an XML object.

lib/zss.x
```diff
@@ -0,0 +1,256 @@
...
+ IMPORT CODE,'zssServer','respondWithUnixFileNotFound'                         
+ IMPORT CODE,'zssServer','respondWithXmlPrinter'                               
+ IMPORT CODE,'zssServer','runServiceThread'                                    
+ IMPORT CODE,'zssServer','safeRealloc'                                         
+ IMPORT CODE,'zssServer','serveFile'                                           
+ IMPORT CODE,'zssServer','serveSimpleTemplate'                                 
+ IMPORT CODE,'zssServer','setConfiguredProperty'                               
+ IMPORT CODE,'zssServer','setContentType'                                      
+ IMPORT CODE,'zssServer','setDefaultJSONRESTHeaders'                           
+ IMPORT CODE,'zssServer','setHttpAuthTrace'
```
1000TurquoisePogs: I'd blacklist all the tracers... doesn't seem like something that plugins should be touching.

lib/zss.x
```diff
@@ -0,0 +1,256 @@
...
+ IMPORT CODE,'zssServer','parseURLMask'                                        
+ IMPORT CODE,'zssServer','parseXMLNode'                                        
+ IMPORT CODE,'zssServer','pluginTypeFromString'                                
+ IMPORT CODE,'zssServer','pluginTypeString'                                    
+ IMPORT CODE,'zssServer','pprintNode'                                          
+ IMPORT CODE,'zssServer','pprintNode2'                                         
+ IMPORT CODE,'zssServer','printStderr'                                         
+ IMPORT CODE,'zssServer','printStdout'                                         
+ IMPORT CODE,'zssServer','printXMLToken'                                       
+ IMPORT CODE,'zssServer','processHttpFragment'
```
1000TurquoisePogs: This seems too internal

lib/zss.x
```diff
@@ -0,0 +1,256 @@
...
+ IMPORT CODE,'zssServer','jsonStringProperty'                                  
+ IMPORT CODE,'zssServer','jsonVerifyHomogeneity'                               
+ IMPORT CODE,'zssServer','logConfigureComponent'                               
+ IMPORT CODE,'zssServer','logConfigureDestination'                             
+ IMPORT CODE,'zssServer','logConfigureDestination2'                            
+ IMPORT CODE,'zssServer','logConfigureStandardDestinations'                    
+ IMPORT CODE,'zssServer','logGetLevel'                                         
+ IMPORT CODE,'zssServer','logSetLevel'                                         
+ IMPORT CODE,'zssServer','logShouldTraceInternal'                              
+ IMPORT CODE,'zssServer','mainHttpLoop'
```
1000TurquoisePogs: Wouldn't be used by a plugin.

PR #36, opened 2019-05-20
Renamed directory for consistency

hogstrom:
Signed-off-by: Matt Hogstrom <matt@hogstrom.org>


hogstrom:
Renamed directory from dco_signoffs to dco-signoffs which is used in other repos for consistency.

PR #41, opened 2019-05-30
Decode an encoded DSN for unusual pds member names

1000TurquoisePogs:
In order for browsers to send requests for pds members with characters like # they need to do uri encoding, which means our code needs to uri decode before interpreting.
Related: https://github.com/zowe/zlux-app-manager/pull/110


gadyck:
Personally, I think you did this in the wrong place.  MFA has the same problem, but not for a dataset.  The decode of the URI needs to be done early, *before* it is even parsed.

1000TurquoisePogs:
Good point, the server core should be doing this so that services dont need to. I'll open an issue on that point.

PR #46, opened 2019-06-05
Updated commit hash for dataset-eof-bug

chcaron01:
https://github.com/zowe/zowe-common-c/pull/40


chcaron01:
@1000TurquoisePogs @jordanfilteau1995

PR #47, opened 2019-06-05
Bugfix symbolic link info

elizabethkirschner:
Added function to not follow symbolic link


elizabethkirschner:
@jordanfilteau1995

PR #49, opened 2019-06-07
Bugfix forbidden returns 200 OK

elizabethkirschner:
Fixed status and error message for forbidden directories.


elizabethkirschner:
@jordanfilteau1995

PR #53, opened 2019-06-28
Adds dynamic logging functionality for plugins

elizabethkirschner:
Added new function to zss.c that creates logging id's for plugins that have specified logging levels.  Id's are stored in a hashtable in the httpserver struct.  The hashtable uses the plugin identifier as the key and stores a struct for the value that contains the created logging id and the log level.

### DIFF
--- a/c/zss.c
+++ b/c/zss.c
@@ -561,11 +561,11 @@ static void installWebPluginDefintionsService(WebPluginListElt *webPlugins, Http
   zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG2, "end %s\n", __FUNCTION__);
 }
 
-static void createLoggingIdentifier(char *identifier, hashtable *pluginLoggingTable,int level,int i) {
+static void createLoggingIdentifier(char *identifier, hashtable *pluginLoggingTable,int level,int idMultiplier) {
   PluginLoggingData *newData = (PluginLoggingData*) safeMalloc(sizeof(PluginLoggingData),"PluginLoggingData");
   if(newData != NULL){
     newData->level = level;
-    uint64 newId = LOG_PROD_PLUGINS + (0x10000 * i); // creates unique logging id
+    uint64 newId = LOG_PROD_PLUGINS + (0x10000 * idMultiplier); // creates unique logging id
     newData->id = newId;
     htPut(pluginLoggingTable, identifier, newData);
     logConfigureComponent(NULL, newId, identifier, LOG_DEST_PRINTF_STDOUT, level);
--- a/c/zss.c
+++ b/c/zss.c
@@ -90,7 +90,8 @@ static int traceLevel = 0;
 static int stringEndsWith(char *s, char *suffix);
 static void dumpJson(Json *json);
 static JsonObject *readPluginDefinition(ShortLivedHeap *slh, char *pluginIdentifier, char *pluginLocation);
-static WebPluginListElt* readWebPluginDefinitions(HttpServer* server, ShortLivedHeap *slh, char *dirname);
+static WebPluginListElt* readWebPluginDefinitions(HttpServer* server, ShortLivedHeap *slh, char *dirname,
+                                                  unsigned int idMultiplier);
 static JsonObject *readServerSettings(ShortLivedHeap *slh, const char *filename);
 static InternalAPIMap *makeInternalAPIMap();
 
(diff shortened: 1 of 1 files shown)


elizabethkirschner:
@jordanfilteau1995

c/zss.c
```diff
@@ -93,6 +93,26 @@ static JsonObject *readPluginDefinition(ShortLivedHeap *slh, char *pluginIdentif
 static WebPluginListElt* readWebPluginDefinitions(HttpServer* server, ShortLivedHeap *slh, char *dirname);
 static JsonObject *readServerSettings(ShortLivedHeap *slh, const char *filename);
 static InternalAPIMap *makeInternalAPIMap();
+ 
+ //redo
+  static int keyCompare(void *key1, void *key2) {
```
jordanfilteau1995: The comment should be informative or not there at all. I think you just forgot to remove them. Also, you should ask Sean if there is a maximum plugin identifier length. strcmp() is ok, but it would be great to use strncmp() incase for some strange reason your input is not null terminated.

c/zss.c
```diff
@@ -93,6 +93,26 @@ static JsonObject *readPluginDefinition(ShortLivedHeap *slh, char *pluginIdentif
...
+  if (strcmp(key1,key2) == 0){
+    return TRUE;
+  }  
+  return FALSE;
+}
+/*
+// copied from collections
```
jordanfilteau1995: What hashing function are you using now?

c/zss.c
```diff
@@ -560,6 +582,43 @@ static void installWebPluginDefintionsService(WebPluginListElt *webPlugins, Http
...
+    else {
+      zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, "error while parsing %s: %s\n", errorBuffer);
+    }
+    JsonObject *logLevels = jsonObjectGetObject(jsonObject, "logLevels");
+    JsonProperty *property = jsonObjectGetFirstProperty(logLevels);
+    int i = 1;
+    while(property != NULL) {
+      char *key = jsonPropertyGetKey(property); 
+      int value = jsonObjectGetNumber(logLevels, key);
+      loggingData *newData = (loggingData*) malloc(sizeof(loggingData));
```
jordanfilteau1995: Usually we use capitals for struct names. Ex. LoggingData. I’d rename to PluginLoggingData for clarity though.

jordanfilteau1995: You need to use safeMalloc.

jordanfilteau1995:
Also, this cannot be merged until @ifakhrutdinov approves.

c/zss.c
elizabethkirschner: the function already exists in collections.c, so instead of copying it into zss.c i use the one that already exists.

c/zss.c
```diff
@@ -93,6 +93,13 @@ static JsonObject *readPluginDefinition(ShortLivedHeap *slh, char *pluginIdentif
 static WebPluginListElt* readWebPluginDefinitions(HttpServer* server, ShortLivedHeap *slh, char *dirname);
 static JsonObject *readServerSettings(ShortLivedHeap *slh, const char *filename);
 static InternalAPIMap *makeInternalAPIMap();
+ 
+  static int keyCompare(void *key1, void *key2) {
```
ifakhrutdinov: There is ```stringCompare``` in zowe-common-c/c/collection.c. No need to add another function.

c/zss.c
```diff
@@ -560,6 +567,41 @@ static void installWebPluginDefintionsService(WebPluginListElt *webPlugins, Http
...
+  char errorBuffer[512] = { 0 };
+  JsonObject *jsonObject = NULL;
+  
+  Json *json = jsonParseFile(slh, filename, errorBuffer, sizeof(errorBuffer));
+  if (json) {
+    if (jsonIsObject(json)) {
+      jsonObject = jsonAsObject(json);
+    }
+    else {
+      zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, "error while parsing %s: %s\n", errorBuffer);
```
ifakhrutdinov: This will ABEND - the format string has two ```%s``` you pass one string. Same for the other similar calls.

c/zss.c
```diff
@@ -560,6 +567,41 @@ static void installWebPluginDefintionsService(WebPluginListElt *webPlugins, Http
...
+    }
+    JsonObject *logLevels = jsonObjectGetObject(jsonObject, "logLevels");
+    JsonProperty *property = jsonObjectGetFirstProperty(logLevels);
+    int i = 1;
+    while(property != NULL) {
+      char *key = jsonPropertyGetKey(property);
```
ifakhrutdinov: Can this be ```NULL```? If yes, the next call will crash.

c/zss.c
```diff
@@ -560,6 +567,41 @@ static void installWebPluginDefintionsService(WebPluginListElt *webPlugins, Http
...
+    else {
+      zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, "error while parsing %s: %s\n", errorBuffer);
+    }
+    JsonObject *logLevels = jsonObjectGetObject(jsonObject, "logLevels");
+    JsonProperty *property = jsonObjectGetFirstProperty(logLevels);
+    int i = 1;
+    while(property != NULL) {
+      char *key = jsonPropertyGetKey(property); 
+      int value = jsonObjectGetNumber(logLevels, key);
+      PluginLoggingData *newData = (PluginLoggingData*) safeMalloc(sizeof(PluginLoggingData),"PluginLoggingData");
```
ifakhrutdinov: This must be checked for ```NULL```.

c/zss.c
```diff
@@ -560,6 +567,41 @@ static void installWebPluginDefintionsService(WebPluginListElt *webPlugins, Http
   zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG2, "end %s\n", __FUNCTION__);
 }
 
+static void parseLoggingData(HttpServer *server, ShortLivedHeap *slh, const char *filename) {
+
+  char errorBuffer[512] = { 0 };
+  JsonObject *jsonObject = NULL;
+  
+  Json *json = jsonParseFile(slh, filename, errorBuffer, sizeof(errorBuffer));
```
ifakhrutdinov: What JSON are you parsing? What's its format?

c/zss.c
```diff
@@ -889,6 +931,8 @@ int main(int argc, char **argv){
     zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "ZSS server settings: address=%s, port=%d\n", address, port);
     server = makeHttpServer2(base,inetAddress,port,requiredTLSFlag,&returnCode,&reasonCode);
     if (server){
+      server->PluginLoggingTable = htCreate(17, stringHash, keyCompare, NULL, NULL); //creates hash table for zowe log
```
ifakhrutdinov: * You can move this into ```parseLoggingData``` and call it ```loadLoggingData```
* Check ```htCreate``` for NULL

c/zss.c
jordanfilteau1995: It can't be NULL. The server will not start if invalid JSON is present in the zluxserver.json file. Also, if the property is null, this code won't execute. In this case, it won't be null, so I don't think it needs to be checked.

c/zss.c
jordanfilteau1995: It's parsing the zluxserver.json file.

c/zss.c
```diff
@@ -560,7 +561,61 @@ static void installWebPluginDefintionsService(WebPluginListElt *webPlugins, Http
   zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG2, "end %s\n", __FUNCTION__);
 }
 
-static WebPluginListElt* readWebPluginDefinitions(HttpServer *server, ShortLivedHeap *slh, char *dirname) {
+static void createLoggingIdentifier(char *identifier, hashtable *pluginLoggingTable,int level,int i) {
```
jordanfilteau1995: change variable name to something more descriptive than "i".

PR #54, opened 2019-07-01
PDSE member creation/deletion (squashed commits)

chcaron01:
Added routes for creation and deletion of PDS members.


chcaron01:
@ifakhrutdinov @jordanfilteau1995

1000TurquoisePogs:
This feels like it is going down an API path we'd regret. Instead of making another /blahDataset/ API, why not refactor into /datasets/blah as we did for /unixfile

1000TurquoisePogs:
I would also say that we'd quickly regret splitting the API between dataset behavior and member behavior. Internally perhaps we would, but the API shouldnt show it. If we are planning to do one before the other then you could imagine this as a way to go about it:

DELETE /datasets/dsn/my.ds(member)
200: deleted

vs
DELETE /datasets/dsn/my.ds
400: dataset deletion not yet implemented, only member deletion at this time

As for the API at large,

/datasetContents -> GET /datasets/dsn/my.ds(member)
/datasetMetadata/hlq -> GET /datasets/detail/hlq
/datasetMetadata/name/my.query* -> GET /datasets/detail/csi/my.query*
/VSAMdatasetContents -> GET /datasets/vsam/dsn/my.ds

Would request some advice from @gadyck and rob scott on the subject of getting VSAM in here in a way that respects its uniqueness without unnecessarily making it a whole different API

gadyck:
I agree that the route of using a multiple type specific /blahDataset/ URI's for the API is a poor choice.

From a high level, I think VSAM datasets should be handled the same as non-VSAM datasets as much as possible.  Many people don't know that z/OS added support for allocation and deletion of [SMS managed VSAM datasets](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.ieab500/vdsws.htm) using JCL or dynamic allocation requests over a decade ago.

A request such as "POST /datasets/dsn/my.ds body={action: allocate, params: {}}" can easily support SMS managed VSAM datasets as long as you don't insist on micro-managing component attributes, such as different CA/CI freespace values for the index and data components.  Too many people don't understand VSAM, and what really should (and should *not*) be specified when creating a VSAM cluster.  They just copy from an example who's roots originated 40 years ago, when VSAM was a different beast.

A request such as "POST /datasets/dsn/my.ds body={action: define, params: {}}" could be used to invoke IDCAMS DEFINE processing if absolutely required.

PR #56, opened 2019-07-08
Top Secret support for user profile extraction.

jordanfilteau1995:
This pull request adds Top Secret functions that will help in the implementation of a security API.

Support for user profile extraction is also present.

Signed-off-by: Jordan Filteau <jfilteau@rocketsoftware.com>

### DIFF
--- a/c/zis/services/secmgmtUtils.c
+++ b/c/zis/services/secmgmtUtils.c
@@ -10,13 +10,36 @@
   Copyright Contributors to the Zowe Project.
 */
 
-#ifndef ZIS_SERVICES_SECMGMTTSS_H_
-#define ZIS_SERVICES_SECMGMTTSS_H_
+#include <metal/metal.h>
+#include <metal/stddef.h>
+#include <metal/stdio.h>
+#include <metal/string.h>
 
-int zisUserProfilesServiceFunctionTSS(CrossMemoryServerGlobalArea *globalArea,
-                                      CrossMemoryService *service, void *parm);
-                                      
+#ifndef METTLE
+#error Non-metal C code is not supported
 #endif
+
+#include "zowetypes.h"
+#include "alloc.h"
+#include "crossmemory.h"
+#include "radmin.h"
+#include "recovery.h"
+#include "zos.h"
+
+bool getCallerUserID(RadminUserID *caller) {
+
+  ACEE aceeData = {0};
+  ACEE *aceeAddress = NULL;
+  cmGetCallerTaskACEE(&aceeData, &aceeAddress);
+  if (aceeAddress == NULL) {
+    return false;
+  }
+
+  caller->length = aceeData.aceeuser[0];
+  memcpy(caller->value, &aceeData.aceeuser[1], sizeof(caller->value));
+
+  return true;
+}
 
 /*
   This program and the accompanying materials are
--- a/c/zis/services/secmgmtUtils.c
+++ b/c/zis/services/secmgmtUtils.c
@@ -20,13 +20,11 @@
 #endif
 
 #include "zowetypes.h"
-#include "alloc.h"
 #include "crossmemory.h"
 #include "radmin.h"
-#include "recovery.h"
 #include "zos.h"
 
-bool getCallerUserID(RadminUserID *caller) {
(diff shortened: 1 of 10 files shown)


c/zis/services/secmgmttss.c
```diff
@@ -0,0 +1,570 @@
...
+/* A helper function which finds the line where the value where we are looking for starts
+ * and returns a pointer to it. This is useful for when the caller wants to start from
+ * a particular value because they either reached the maximum they could return in a 
+ * request, or if they want to information on a single user. It also returns pointers
+ * to the lastByteOffset and the blockBytesExtracted so when continuing to parse,
+ * we still have the right place through the output.
+ */
+static const RadminVLText *findLineOfStart(const RadminVLText *currentLine, const char *searchValue, 
+                                           int searchValueMaxLength, const char *searchKey,
+                                           const unsigned int *lastByteOffset, unsigned int *blockBytesExtracted,
```
ifakhrutdinov: Why don't you just pass the value instead of the address of ```lastByteOffset```?

ifakhrutdinov: A better name for ```blockBytesExtracted``` would be ```offsetInBlock``` and for ```lastByteOffset``` is ```maxOffsetInBlock```, that will indicate the purpose better.

c/zis/services/tssparsing.c
```diff
@@ -0,0 +1,401 @@
...
+#include <metal/stdio.h>
+#include <metal/string.h>
+
+#ifndef METTLE
+#error Non-metal C code is not supported
+#endif
+
+#include "zowetypes.h"
+#include "alloc.h"
+#include "crossmemory.h"
```
ifakhrutdinov: Why do you need ```crsossmemory``` here?

h/zis/services/tssparsing.h
```diff
@@ -0,0 +1,73 @@
...
+ */
+typedef struct KeyData_t {
+  unsigned int startPos;
+  unsigned int endPos;
+  unsigned int delimPos;
+  char keyName[MAX_KEY_LENGTH];
+  int keyLength;
+} KeyData;
+
+bool isErrorDetected(const char *str, int len);
```
ifakhrutdinov: Can you please prefix these so that there is no name collision? E.g. ```tssutil```

jordanfilteau1995:
@ifakhrutdinov Changes have been made. Please review at your convenience. Thank you.

jordanfilteau1995:
> 
> 
> My other concerns:
> 
>     * The performance will be potentially bad when we use `startUserID` because we have to reach it manually
> 
>     * What if the start user gets removed or not found? It seems we won't extract anything

Yeah, there isn't a callable service, so I have to find it manually. There isn't really any counter play for that which I can think of.

c/zis/services/secmgmtUtils.c
```diff
@@ -0,0 +1,52 @@
...
+
+#ifndef METTLE
+#error Non-metal C code is not supported
+#endif
+
+#include "zowetypes.h"
+#include "alloc.h"
+#include "crossmemory.h"
+#include "radmin.h"
+#include "recovery.h"
```
ifakhrutdinov: What do you need ```recovery.h``` for here?

c/zis/services/secmgmtUtils.c
```diff
@@ -0,0 +1,52 @@
...
+#include "zos.h"
+
+bool getCallerUserID(RadminUserID *caller) {
```
ifakhrutdinov: Please use a prefix. For example, ```secmgmtGetCallerUserID``` or ```secutilGetCallerUserID```

c/zis/services/secmgmtUtils.c
```diff
@@ -0,0 +1,52 @@
...
+
+/*
+  This program and the accompanying materials are
+  made available under the terms of the Eclipse Public License v2.0 which accompanies
+  this distribution, and is available at https://www.eclipse.org/legal/epl-v20.html
+
+  SPDX-License-Identifier: EPL-2.0
+
+  Copyright Contributors to the Zowe Project.
+*/
```
ifakhrutdinov: Put a newline here.

PR #64, opened 2019-07-30
Check if volumes are available when using the dataset API

dmitrynikolaev:
Signed-off-by: Dmitry Nikolaev <dnikolaev@rocketsoftware.com>


dmitrynikolaev:
Due to unclear Jenkins errors, it fails to build the PR

PR #79, opened 2019-08-20
RSJWT support

fkovinAtRocket:
Depends on https://github.com/zowe/zowe-common-c/pull/73


bin/zss-import-jwt-key.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/sh
```
1000TurquoisePogs: This will need the open source header

PR #81, opened 2019-08-28
Server REST API

timgerstel:
REST API which enables administrators to query the status of ZSS server.

/server/agent -> list routes
/server/agent/config -> display server config in the form of zluxserver.json with respect to CLI args.
/server/agent/log -> display server log file
/server/agent/logLevels -> display logger verbosity levels
/server/agent/environment -> display information about the environment in which the ZSS server is running.

Do not merge without: 
https://github.com/zowe/zowe-common-c/pull/81


c/restService.c
```diff
@@ -0,0 +1,266 @@
...
+    if (!strcmp(request->method, methodGET)) {
+      char *l1 = stringListPrint(request->parsedFile, 2, 1, "/", 0);
+      if(!strcmp(l1, "")){
+        respondWithServerRoutes(response);
+      }
+      else if (!strcmp(l1, "config")){
+        respondWithServerConfig(response);
+      }
+      else if (!strcmp(l1, "log")) {
+        if(strcmp(getenv("ZSS_LOG_FILE"), "")){
```
ifakhrutdinov: ```getenv``` calls may return NULL pointers. Please check for NULL before using the results of ```getenv```.

c/restService.c
```diff
@@ -0,0 +1,266 @@
...
+
+static int serveStatus(HttpService *service, HttpResponse *response) {
+  HttpRequest *request = response->request;
+  JsonObject *dataserviceAuth = jsonObjectGetObject(serverConfig, "dataserviceAuthentication");
+  int rbacParm = jsonObjectGetBoolean(dataserviceAuth, "rbac");
+  if(!rbacParm){
+     respondWithError(response, HTTP_STATUS_UNAUTHORIZED, "Unauthorized - RBAC is disabled.  Enable in zluxserver.json");
+  } else {
+    if (!strcmp(request->method, methodGET)) {
+      char *l1 = stringListPrint(request->parsedFile, 2, 1, "/", 0);
```
ifakhrutdinov: Can ```l1``` be NULL?

c/restService.c
```diff
@@ -0,0 +1,266 @@
...
+  jsonStart(out);
+  jsonAddString(out, "logDirectory", getenv("ZSS_LOG_FILE"));
+  jsonAddString(out, "agentName", "zss");
+  jsonAddString(out, "agentVersion", productVersion);
+  jsonAddString(out, "arch", unameRet.sysname);
+  jsonAddString(out, "osRelease", unameRet.release);
+  jsonAddString(out, "hardwareIdentifier", unameRet.machine);
+  jsonAddString(out, "hostname", unameRet.nodename);
+  jsonStartObject(out, "userEnvironment");
+  char *env_var = strdup(*environ);
```
ifakhrutdinov: According to the doc, ```strdup``` may return NULL.

c/restService.c
```diff
@@ -0,0 +1,266 @@
...
+  int i = 1;
+  for (; env_var; i++) {
+    char *var_name = strtok(env_var, "=");
+    char *var_value = strtok(NULL, "=");
+    jsonAddString(out, var_name, var_value);
+    env_var = strdup(*(environ+i));
```
ifakhrutdinov: What if ```*(environ+i) == NULL```, shouldn't that be checked for NULL in the for loop instead of ```env_var```?

c/restService.c
```diff
@@ -0,0 +1,266 @@
...
+  jsonAddString(out, "agentVersion", productVersion);
+  jsonAddString(out, "arch", unameRet.sysname);
+  jsonAddString(out, "osRelease", unameRet.release);
+  jsonAddString(out, "hardwareIdentifier", unameRet.machine);
+  jsonAddString(out, "hostname", unameRet.nodename);
+  jsonStartObject(out, "userEnvironment");
+  char *env_var = strdup(*environ);
+  int i = 1;
+  for (; env_var; i++) {
+    char *var_name = strtok(env_var, "=");
```
ifakhrutdinov: What if there is no ```"="``` in the string?

c/restService.c
```diff
@@ -0,0 +1,266 @@
...
+int installServerStatusService(HttpServer *server, JsonObject *serverSettings, char* productVer)
+{
+  HttpService *httpService = makeGeneratedService("REST_Service", "/server/agent/**");
+  httpService->authType = SERVICE_AUTH_NATIVE_WITH_SESSION_TOKEN;
+  httpService->serviceFunction = &serveStatus;
+  httpService->runInSubtask = TRUE;
+  httpService->doImpersonation = TRUE;
+  registerHttpService(server, httpService);
+  serverConfig = serverSettings;
+  memcpy(productVersion, productVer, 40);
```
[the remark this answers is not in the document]
timgerstel: I should verify that it is null terminated but this line is how I know it will be equal to 40 bytes: https://github.com/zowe/zss/blob/master/c/zss.c#L84

c/serverStatusService.c
```diff
@@ -0,0 +1,266 @@
...
+#include "restService.h"
+#pragma linkage (EXSMFI,OS)
+
+#ifdef __ZOWE_OS_ZOS
+static int serveStatus(HttpService *service, HttpResponse *response);
+static JsonObject* serverConfig = NULL;
+char productVersion[40];
+extern char **environ;
+
+typedef int EXSMFI();
```
fkovinAtRocket: the correct syntax for a function with no arguments is` EXSMFI(void)`. `EXSMFI()` means "a function which you can all with any number of arguments, all of them will be cast to int"

c/serverStatusService.c
```diff
@@ -0,0 +1,266 @@
...
+               &subtype,
+               buffer,
+               &bufferlen,
+               &cpu_util,
+               &demand_paging,
+               &options,
+               &mvs_srm,
+               &zaap_util,
+               &ziip_util);
+  sprintf(pid, "%d", getpid());
```
fkovinAtRocket: I would instead to 
`snprintf(pid, sizeof (pid), "%d", getpid());`
etc

c/serverStatusService.c
```diff
@@ -0,0 +1,266 @@
...
+}
+
+void respondWithServerEnvironment(HttpResponse *response){
+  /*Information about parameters for smc_func: https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.erbb700/smfp.htm#smfp*/
+  jsonPrinter *out = respondWithJsonPrinter(response);
+  struct utsname unameRet;
+  uname(&unameRet);
+  char pid[64];
+  char ppid[64];
+  char dp[64];
```
fkovinAtRocket: It's unclear from the context what `dp`, `mvs_u`, etc, are. Can you please add a little more descriptive names?

c/serverStatusService.c
```diff
@@ -0,0 +1,266 @@
...
+  char cpu_u[64];
+  char mvs_u[64];
+  char ziip_u[64];
+  char zaap_u[64];
+  char* buffer;
```
fkovinAtRocket: We ignore the variable after the call. Is that intentional?

c/serverStatusService.c
```diff
@@ -0,0 +1,266 @@
...
+  char *env_var = strdup(*environ);
+  int i = 1;
+  for (; env_var; i++) {
+    char *var_name = strtok(env_var, "=");
+    char *var_value = strtok(NULL, "=");
+    jsonAddString(out, var_name, var_value);
+    env_var = strdup(*(environ+i));
+  }
+  jsonEndObject(out);
+  jsonAddString(out, "demandPagingRate", dp);
```
fkovinAtRocket: Why don't we do this here instead:
```
chat smfVarBuffer[64];

snprintf(smfVarBuffer, sizeof (smfVarBuffer), "%d", demand_paging);
jsonAddString(out, "demandPagingRate", dp); 

snprintf(smfVarBuffer, sizeof (smfVarBuffer), "%d", cpu_util);
jsonAddString(out, "stdCP_CPU_Util", dp); 
etc...
```

c/serverStatusService.c
```diff
@@ -0,0 +1,266 @@
...
+  char mvs_u[64];
+  char ziip_u[64];
+  char zaap_u[64];
+  char* buffer;
+  int rc = 0;
+  int reqtype = 0x00000005; //fullword. request type
+  int rectype = 0x0000004F; //SMF record type, only type 79 is supported
+  int subtype = 0x00000009; //SMF record subtype
+  int bufferlen = 247488; /* length of SMF record buffer */
+  int cpu_util = 0x00000000;
```
fkovinAtRocket: naming: cpuUtil, demandPaging, etc.

c/serverStatusService.c
```diff
@@ -0,0 +1,266 @@
...
+  jsonEndObject(out);
+  jsonAddString(out, "demandPagingRate", dp); 
+  jsonAddString(out, "stdCP_CPU_Util", cpu_u);
+  jsonAddString(out, "stdCP_MVS_SRM_CPU_Util", mvs_u);
+  jsonAddString(out, "ZAAP_CPU_Util", zaap_u);
+  jsonAddString(out, "ZIIP_CPU_Util", ziip_u);
+  jsonAddString(out, "PID", pid);
+  jsonAddString(out, "PPID", ppid);
+  jsonEnd(out);
+  safeFree(env_var, sizeof(*environ)+1);
```
fkovinAtRocket: `*evniron` is just a pointer, so `sizeof(*environ)` is not the length of the string (although we don't care about that in LE, this is still misleading)

c/serverStatusService.c
```diff
@@ -0,0 +1,266 @@
...
+        respondWithLogLevels(response);
+      }
+      else if (!strcmp(l1, "environment")) {
+        respondWithServerEnvironment(response);
+      }
+      else {
+        respondWithJsonError(response, "Invalid path", 400, "Bad Request");
+      }
+    }
+    else{
```
fkovinAtRocket: formatting: please remove the newline before `else`

c/serverStatusService.c
```diff
@@ -0,0 +1,266 @@
...
+  jsonEnd(out);
+  safeFree(env_var, sizeof(*environ)+1);
+  finishResponse(response);
+}
+
+static int serveStatus(HttpService *service, HttpResponse *response) {
+  HttpRequest *request = response->request;
+  JsonObject *dataserviceAuth = jsonObjectGetObject(serverConfig, "dataserviceAuthentication");
+  int rbacParm = jsonObjectGetBoolean(dataserviceAuth, "rbac");
+  if(!rbacParm){
```
fkovinAtRocket: The name of the variable and the config JSON field is confusing. How is this related to RBAC?

c/serverStatusService.c
```diff
@@ -0,0 +1,266 @@
...
+  int cpu_util = 0x00000000;
+  int demand_paging = 0x00000000;
+  int options = 0x00000000;
+  int mvs_srm = 0x00000000;
+  int zaap_util = 0x00000000;
+  int ziip_util = 0x00000000;
+  buffer = (char *)safeMalloc(bufferlen, "buffer");
+  memset(buffer, 0, bufferlen);
+  smf_func = (EXSMFI *)fetch("ERBSMFI");
+  rc = (*smf_func)(&reqtype,
```
fkovinAtRocket: can `rc` indicate an error?

c/serverStatusService.c
```diff
@@ -0,0 +1,266 @@
...
+  jsonStart(out);
+  jsonAddString(out, "logDirectory", getenv("ZSS_LOG_FILE"));
+  jsonAddString(out, "agentName", "zss");
+  jsonAddString(out, "agentVersion", productVersion);
+  jsonAddString(out, "arch", unameRet.sysname);
+  jsonAddString(out, "osRelease", unameRet.release);
+  jsonAddString(out, "hardwareIdentifier", unameRet.machine);
+  jsonAddString(out, "hostname", unameRet.nodename);
+  jsonStartObject(out, "userEnvironment");
+  char *env_var = strdup(*environ);
```
fkovinAtRocket: Like Irek said above, null check

c/serverStatusService.c
timgerstel: The member that states whether RBAC is enabled is under the dataserviceAuthentication member.
```
dataserviceAuthentication{
  "defaultAuth...": "...",
  "rbac": true
  ....
}
```

timgerstel:
@ifakhrutdinov @fkovinAtRocket Thank you guys very much for the verbose feedback, I appreciate you guys taking time to review this for me.  I have implemented all of the requested changes, if one or both of you wouldn't mind taking a look at the changes I would be very grateful!

c/restService.c
timgerstel: I do not believe this is a possible scenario for environment variables.  There must be at a minimum one "=" unless I am mistaken

c/restService.c
timgerstel: Doesn't the end condition "env_var;" mean "loop until env_var is null"?  In which case, when *(environ+i) == NULL, env_var will also be null and thus exit the loop.

c/restService.c
timgerstel: I do not believe stringListPrint() can return null.

c/zss.c
```diff
@@ -817,7 +818,7 @@ static int validateFilePermissions(const char *filePath) {
   } else {
     return validateConfigPermissionsInner(filePath);
   }
-}
+}
```
ifakhrutdinov: Please don't commit trailing spaces.

c/serverStatusService.c
```diff
@@ -0,0 +1,262 @@
...
+      break;
+    }
+    char *name = safeMalloc(strlen(env_var)-strlen(len) + 1, "env_var name");
+    snprintf(name, sizeof(char)*(strlen(env_var)-strlen(len) + 1), "%s", env_var);
+    jsonAddString(out, name, len+1);
+    safeFree(name, strlen(name));
+    env_var = *(environ+i);
+  }
+  jsonEndObject(out);
+  snprintf(smfVarBuffer, sizeof(smfVarBuffer), "%d%%", demandPaging);
```
ifakhrutdinov: Why do we need to convert integers to strings? Client of this service now have to parse these values (and take into account the fact that some values have a percent sign).

Can we do ```jsonAddInt```?

c/serverStatusService.c
```diff
@@ -0,0 +1,262 @@
...
+               buffer,
+               &bufferlen,
+               &cpuUtil,
+               &demandPaging,
+               &options,
+               &mvsSrm,
+               &zaapUtil,
+               &ziipUtil);
+  if(rc > 0){
+    respondWithError(response, HTTP_STATUS_BAD_REQUEST, "Unable to fetch from RMF data interface service");
```
fkovinAtRocket: and then free the buffer and return, right?

c/serverStatusService.c
fkovinAtRocket: It's just not obvious how this is related to the rest of the code. Why would we not send someone a status response if rbac is disabled in the config?

c/serverStatusService.c
timgerstel: I free the buffer above line 156 now, its never used after and does not contain (human) readable data.  respondWithError returns the response.

timgerstel:
> 
> 
> @timgerstel Please see my comments. I haven't finished reviewing yet, will look at the code more later.

Thank you again, I have implemented your requested changes.

c/restService.c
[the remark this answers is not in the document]
1000TurquoisePogs: This can be best solved by using a `#DEFINE` to state that both are 40.

c/serverStatusService.c
1000TurquoisePogs: We need to send a response when RBAC is disabled, but it should be an HTTP 400 bad request. This service is conditional on RBAC being enabled, because it is a sensitive URL.

1000TurquoisePogs:
Also, dont forget DCO as it is challenging to fix the longer you wait.

c/serverStatusService.c
fkovinAtRocket: but we still need to return from the function now

c/serverStatusService.c
fkovinAtRocket: That would be a comment I was looking for :-)

PR #82, opened 2019-09-05
Server REST API

timgerstel:
REST API which enables administrators to query the status of ZSS server.

/server/agent -> list routes
/server/agent/config -> display server config in the form of zluxserver.json with respect to CLI args.
/server/agent/log -> display server log file
/server/agent/logLevels -> display logger verbosity levels
/server/agent/environment -> display information about the environment in which the ZSS server is running.

Do not merge without:
zowe/zowe-common-c#81


timgerstel:
Hello @ifakhrutdinov @fkovinAtRocket, I messed up the other branch trying to sign off an old commit.  Here is a new pull request, with your requested changes implemented.

ifakhrutdinov:
@timgerstel that's unfortunate, was it impossible to fix the branch? The problem is, we don't have all the comments from PR #81.

timgerstel:
> 
> 
> @timgerstel that's unfortunate, was it impossible to fix the branch? The problem is, we don't have all the comments from PR #81.

@ifakhrutdinov I will try to fix it again since this branch exists now with functional code.  I was hesitant to create this new PR for that exact reason (missing comments).

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+#include "utils.h"
+#include "socketmgmt.h"
+
+#include "logging.h"
+#include "zssLogging.h"
+#include "serverStatusService.h"
+
+#ifdef __ZOWE_OS_ZOS
+static int serveStatus(HttpService *service, HttpResponse *response);
+extern char **environ;
```
jordanfilteau1995: Should this really be a global?

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+
+typedef int EXSMFI(int *reqType, int *recType, int *subType,
+                   char* buffer, int *bufferLen, int *cpuUtil,
+                   int *dpRate, int *options, int *mvs, int *zaap, int *ziip);
+EXSMFI *smfFunc;
+
+int installServerStatusService(HttpServer *server, JsonObject *serverSettings, char* productVer)
```
jordanfilteau1995: Can this be void? You return zero regardless, since there is no failure condition.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+
+void respondWithServerEnvironment(HttpResponse *response, ServerAgentContext *context){
+  /*Information about parameters for smf_unc: https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.erbb700/smfp.htm#smfp*/
+  struct utsname unameRet;
+  char hostnameBuffer[256];
+  char* buffer;
+  time_t curr_time;
+  char* c_time_str;
+  int rc = 0;
+  int reqtype = 0x00000005; //fullword. request type
```
jordanfilteau1995: If these are meant to be constant, then you should DEFINE them all somewhere.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+    jsonPrintObject(out, logLevels);
+  }
+  jsonEnd(out);
+  finishResponse(response);
+}
+
+void respondWithServerEnvironment(HttpResponse *response, ServerAgentContext *context){
+  /*Information about parameters for smf_unc: https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.erbb700/smfp.htm#smfp*/
+  struct utsname unameRet;
+  char hostnameBuffer[256];
```
jordanfilteau1995: You should try not to just use arbitrary numbers for the size of a character array. It would be better to define that somewhere, but this is a minor and stylistic concern.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+  char* buffer;
+  time_t curr_time;
```
jordanfilteau1995: You still have these odd duck variable names.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+  int rectype = 0x0000004F; //SMF record type, only type 79 is supported
+  int subtype = 0x00000009; //SMF record subtype
+  int bufferlen = 716800; // Yes the SMF buffer is actually 700Kb roughly
+  int cpuUtil = 0x00000000;
+  int demandPaging = 0x00000000;
+  int options = 0x00000000;
+  int mvsSrm = 0x00000000;
+  int zaapUtil = 0x00000000;
+  int ziipUtil = 0x00000000;
+  uname(&unameRet);
```
jordanfilteau1995: You could really just implement BPXUNA and make this compatible with LE and METAL. @ifakhrutdinov what do you think?

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+  gethostname(hostnameBuffer, sizeof(hostnameBuffer));
```
jordanfilteau1995: Same here, where you could implement BPXHST.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+               buffer,
+               &bufferlen,
+               &cpuUtil,
+               &demandPaging,
+               &options,
+               &mvsSrm,
+               &zaapUtil,
+               &ziipUtil);
+  safeFree(buffer, bufferlen);
+  curr_time = time(NULL);
```
jordanfilteau1995: You could use the functions we have in our library (timeutl.c), which would allow this to be METAL compatible.

c/serverStatusService.c
jordanfilteau1995: Nevermind, you need to pass these by address.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+  c_time_str = ctime(&curr_time);
+  if(c_time_str[strlen(c_time_str) - 1] == '\n'){
```
jordanfilteau1995: I don't like using a strlen() for the size of a character array. If what you are operating on isn't null terminated, you are going to run into problems.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+  jsonAddString(out, "agentName", "zss");
+  jsonAddString(out, "agentVersion", context->productVersion);
+  jsonAddString(out, "arch", unameRet.sysname);
+  jsonAddString(out, "osRelease", unameRet.release);
+  jsonAddString(out, "osVersion", unameRet.version);
+  jsonAddString(out, "hardwareIdentifier", unameRet.machine);
+  jsonAddString(out, "hostname", hostnameBuffer);
+  jsonAddString(out, "nodename", unameRet.nodename);
+  jsonStartObject(out, "userEnvironment");
+  char *env_var = *environ;
```
jordanfilteau1995: This isn't camel case.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+  for (int i = 1; env_var; i++) {
+    char *equalSign = strchr(env_var, '=');
```
jordanfilteau1995: You could just use indexOfChar (?), which we have in our library and not have to use a loop.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+    if(equalSign == NULL){
+      break;
+    }
+    int nameLen = strlen(env_var) - strlen(equalSign);
```
jordanfilteau1995: If you could find the maximum length of this on z/OS, then you could allocate it on the stack instead.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+    env_var = *(environ+i);
+  }
+  jsonEndObject(out);
+  jsonAddInt(out, "demandPagingRate", demandPaging);
+  jsonAddInt(out, "stdCP_CPU_Util", cpuUtil);
+  jsonAddInt(out, "stdCP_MVS_SRM_CPU_Util", mvsSrm);
+  jsonAddInt(out, "ZAAP_CPU_Util", zaapUtil);
+  jsonAddInt(out, "ZIIP_CPU_Util", ziipUtil);
+  jsonAddInt(out, "PID", getpid());
+  jsonAddInt(out, "PPID", getppid());
```
jordanfilteau1995: You could use BPX1GPP here.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+    safeFree(name, nameBufferLen);
+    env_var = *(environ+i);
+  }
+  jsonEndObject(out);
+  jsonAddInt(out, "demandPagingRate", demandPaging);
+  jsonAddInt(out, "stdCP_CPU_Util", cpuUtil);
+  jsonAddInt(out, "stdCP_MVS_SRM_CPU_Util", mvsSrm);
+  jsonAddInt(out, "ZAAP_CPU_Util", zaapUtil);
+  jsonAddInt(out, "ZIIP_CPU_Util", ziipUtil);
+  jsonAddInt(out, "PID", getpid());
```
jordanfilteau1995: You could use BPXGPI here.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+  jsonAddInt(out, "PPID", getppid());
+  jsonEnd(out);
+  finishResponse(response);
+}
+
+static int serveStatus(HttpService *service, HttpResponse *response) {
+  HttpRequest *request = response->request;
+  ServerAgentContext *context = service->userPointer;
+  //This service is conditional on RBAC being enabled because it is a sensitive URL that only RBAC authorized users should be able to access
```
jordanfilteau1995: This line is more than what's readable without scrolling on Github. If you go back and make changes, you'd might as well fix this line length.

c/serverStatusService.c
```diff
@@ -0,0 +1,272 @@
...
+    if(!strcmp(request->method, methodGET)){
+      char *l1 = stringListPrint(request->parsedFile, 2, 1, "/", 0);
+      if(!strcmp(l1, "")){
+        respondWithServerRoutes(response);
+      }else if (!strcmp(l1, "config")){
+        respondWithServerConfig(response, context->serverConfig);
+      }else if (!strcmp(l1, "log")) {
+        char* logDir = getenv("ZSS_LOG_FILE");
+        if(logDir == NULL || strcmp(logDir, "")){
+          respondWithUnixFile2(NULL, response, logDir, 0, 0, false);
```
jordanfilteau1995: You can use respondWithUnixFileContents() which chains to call this function eventually.

c/serverStatusService.c
timgerstel: I cannot find a function or service named BPXUNA

c/serverStatusService.c
timgerstel: http://man7.org/linux/man-pages/man2/gethostname.2.html

> "SUSv2 guarantees that "Host names are limited to 255 bytes".  POSIX.1
>        guarantees that "Host names (not including the terminating null byte)
>        are limited to HOST_NAME_MAX bytes".  On Linux, HOST_NAME_MAX is
>        defined with the value 64, which has been the limit since Linux 1.0
>        (earlier kernels imposed a limit of 8 bytes)."

256 wasnt arbitrary but I will make it a define

c/serverStatusService.c
ifakhrutdinov: We need to think how to wrap BPX calls. If @timgerstel does that, we'll need a longer discussion. I think we should do that later. ```uname``` is probably BPX1UNA under the covers.

c/serverStatusService.c
timgerstel: respondWithUnixFileContents() can't be used because it requires a query to unset asB64 in respondWithUnixFile2().

httpserver.c line 3536:
```
    bool asB64 = TRUE;
    char *base64Param = getQueryParam(response->request, "responseType");

    if (NULL != base64Param) {
      if (!strcmp(strupcase(base64Param), "RAW")) {
        asB64 = FALSE;
      }
    }

    respondWithUnixFile2(service, response, absolutePath, jsonMode, autocvt, asB64);
```

c/serverStatusService.c
jordanfilteau1995: Yeah, it is.

c/serverStatusService.c
jordanfilteau1995: OK

h/serverStatusService.h
```diff
@@ -0,0 +1,38 @@
...
+  
+  SPDX-License-Identifier: EPL-2.0
+  
+  Copyright Contributors to the Zowe Project.
+*/
+
+#ifndef __SERVER_STATUS_SERVICE_H__
+#define __SERVER_STATUS_SERVICE_H__
+#include "json.h"
+#include "httpserver.h"
```
jordanfilteau1995: Put in C file if they aren't needed here.

c/serverStatusService.c
jordanfilteau1995: Yeah. I guess we can go back and deal with it after.

c/serverStatusService.c
```diff
@@ -0,0 +1,284 @@
...
+  } else {
+    if(!strcmp(request->method, methodGET)){
+      char *l1 = stringListPrint(request->parsedFile, 2, 1, "/", 0);
+      if(!strcmp(l1, "")){
+        respondWithServerRoutes(response);
+      }else if (!strcmp(l1, "config")){
+        respondWithServerConfig(response, context->serverConfig);
+      }else if (!strcmp(l1, "log")) {
+        char* logDir = getenv("ZSS_LOG_FILE");
+        if(logDir == NULL || strcmp(logDir, "")){
```
ifakhrutdinov: Seems like a bug. Did you mean the following?
```c
if (logDir != NULL && strcmp(logDir, "")) {
```

c/serverStatusService.c
```diff
@@ -0,0 +1,284 @@
...
+  JsonObject *dataserviceAuth = jsonObjectGetObject(context->serverConfig, "dataserviceAuthentication");
+  int rbacParm = jsonObjectGetBoolean(dataserviceAuth, "rbac");
+  if(!rbacParm){
+     respondWithError(response, HTTP_STATUS_BAD_REQUEST, "Set dataserviceAuthentication.rbac to true in server configuration");
+  } else {
+    if(!strcmp(request->method, methodGET)){
+      char *l1 = stringListPrint(request->parsedFile, 2, 1, "/", 0);
+      if(!strcmp(l1, "")){
+        respondWithServerRoutes(response);
+      }else if (!strcmp(l1, "config")){
```
ifakhrutdinov: I think we should be more consistent with formatting. Please use spaces after in argument lists and before/after brackets.

For example,
```c
if(!strcmp(l1, "")){
}else if (!strcmp(l1, "config")){
 }else if (!strcmp(l1, "log")) {
} else {
```
should be
```c
if (!strcmp(l1, "")) {
} else if (!strcmp(l1, "config")) {
 } else if (!strcmp(l1, "log")) {
} else {
```

c/serverStatusService.c
daveyc: You can flatten these kind of bugs by using macros or (preferably) inline functions instead of ```strcmp()``` for checking string equality. I prefer using bool functions as I dislike using boolean logic for ```strcmp()``` which is almost always used for equality;

````
inline bool strne(const char * lhs, const char * rhs) { 
  return lhs != NULL && rhs != NULL && strcmp(lhs, rhs) != 0;
}
````

Comment (timgerstel) [reply c/serverStatusService.c]: I quite like this inline function, I think I will implement it!

Comment (daveyc) [reply c/serverStatusService.c]: For that particular case you may prefer a different function.

````c
inline bool str_is_empty(const char * str) {
  return str ==  NULL || str[0] == '\0';
}
````
These kind of string functions are as old as the hills. C has a poor API for checking string equality which necessitates writing wrappers. You can see examples in lots of [open source software](https://github.com/systemd/systemd/blob/master/src/basic/string-util.h).

File: c/serverStatusService.c [hunk 20]
```diff
@@ -0,0 +1,297 @@
...
+  } else {
+    if (!strcmp(request->method, methodGET)) {
+      char *l1 = stringListPrint(request->parsedFile, 2, 1, "/", 0);
+      if (!strcmp(l1, "")) {
+        return respondWithServerRoutes(response);
+      } else if (!strcmp(l1, "config")) {
+        return respondWithServerConfig(response, context->serverConfig);
+      } else if (!strcmp(l1, "log")) {
+        char* logDir = getenv("ZSS_LOG_FILE");
+        if(strne(logDir, "")){
```
ifakhrutdinov: I suggest you configure your IDE to automatically format this kind of stuff.

ifakhrutdinov:
@timgerstel please squash the commits before the merge. Please also add the description of changes to the commit message of the final commit.

PR #96, opened 2019-10-17
Add scripts to deploy ZIS-related load libraries and update ZIS PARMLIB.

ifakhrutdinov:
Changes:
* A new script has been added. It allows users to specify a SAMPLIB file with which they want to update their ZIS PARMLIB member.
* Adjust the existing ZIS plugin install script to have an "extended install" option for LOADLIB and SAMPLIB deployment.

Using the extended install option:
* Use have a directory with load libraries for ZIS (multiple plugins) called ```dev_load```
* Use have a directory with sample files for ZIS called ```dev_sample```
* Issue the following command:
```
./zis-plugin-install.sh extended-install ./dev_load ./dev_sample
```
* The script will copy the load libraries to the ZIS LOADLIB
* The script will update the ZIS PARMLIB member with the content of the sample files in ```dev_sample```
  * If a key-value pair exists in the ZIS PARMLIB member, the script will update the value
  * If a key-value pair doesn't exist in the ZIS PARMLIB member, the script will append it

Additional features:
* If a key-value pair in your sample file is defined using the shell variable syntax (e.g. ZWES.FOO.VERSION=$VER), the script will dereference environmental variable $VER and use that value to update the ZIS PARMLIB member

### DIFF
--- a/bin/zis-parmlib-update.sh
+++ b/bin/zis-parmlib-update.sh
@@ -1,0 +1,284 @@
+#!/bin/sh
+# This program and the accompanying materials are
+# made available under the terms of the Eclipse Public License v2.0 which accompanies
+# this distribution, and is available at https://www.eclipse.org/legal/epl-v20.html
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Copyright Contributors to the Zowe Project.
+
+# This script adds or updates pramaters in ZIS PARMLIB member specified in
+# the provided SAMPLIB file.
+
+# Make sure all files created in this script are only accesible by the owner.
+umask go-rwx
+
+# Prohibit undefined variables.
+set -u
+
+# ZIS-related global variables.
+ZIS_PREFIX=${ZIS_PREFIX:-$USER.DEV}
+ZIS_PARMLIB=${ZIS_PARMLIB:-$ZIS_PREFIX.PARMLIB}
+ZIS_PARMLIB_MEMBER=${ZIS_PARMLIB_MEMBER:-ZWESIP00}
+
+#################################################
+# Generate a name for a temporary file in /tmp
+#
+# Globals:
+#   RANDOM
+# Arguments:
+#   None
+# Returns:
+#   The generated file name
+#################################################
+make_temp_filename() {
+  name=/tmp/zis-sample-install-$$-$RANDOM
+  while [ -e $name ]
+  do
+    name=/tmp/zis-sample-install-$$-$RANDOM
+  done
+  echo $name
+}
+
+#################################################
+# Try to resolve values that are defined using
+# environmental variables, otherwise return
+# the original value
+#
+# Globals:
+#   None
+# Arguments:
+#   $1 Value from a key-value pair
+# Returns:
+#   * If an env variable is provided, its value
+#     is returned on success
+#   * If an env varible if provided and
+#     the variable is not defined,
+#     string VALUE_NOT_FOUND is returned
+#   * The original value is returned
+#################################################
+resolve_env_parameter() {
+
+  parm=$1
+
+  # Are we dealing with an env variable based value?
+  echo $parm | grep -E "^[$]{1}[A-Z0-9_]+$" 1>/dev/null
+  # Yes, resolve the value using eval, otherwise return the value itself.
+  if [ $? -eq 0 ]; then
+    ( eval "echo $parm" 2>/dev/null )
+    if [ $? -ne 0 ]; then
+      echo "VALUE_NOT_FOUND"
+    fi
+  else
+    echo $parm
+  fi
+
+}
+
+#################################################
+# Add (or update) a key-value pair to a PARMLIB
+# member.
+#
+# Globals:
+#   RANDOM
+# Arguments:
+#   $1 PARMLIB file to be updated
+#   $2 Parameter to be added/updated
+#   $3 Parameter value
+# Returns:
+#   0 on success, 8 on error.
+#################################################
+add_key_value_parm_to_parmlib() {(
+
+  # Input parameters.
+  parm_file=$1
+  parm_key=$2
+  parm_value=$(resolve_env_parameter $3)
+
+  # Check if we recevied a non-empty value for the key (if the value has been
+  # defined using an environmental variable).
+  if [ "$parm_value" = "VALUE_NOT_FOUND" ]; then
+    echo "error: value for $parm_key has not been provided"
+    exit 8
+  fi
+
+  key_value="$parm_key=$parm_value"
+  parm_file_updated="$parm_file"_updated
+
+  # Make a temp file for sed.
+  touch $parm_file_updated
+  if [ $? -ne 0 ]; then
+    exit 8
+  fi
+
+  # Make sure the temporary file produced by sed is removed.
+  trap "rm $parm_file_updated" EXIT
+
+  echo "info: adding/updaing $parm_key with value $parm_value"
+
+  # Try to find the key in our parameter file.
+  grep "^[[:blank:]]*$parm_key=" $parm_file 1>/dev/null
+
+  # Does the key already exist? If yes, subtitute the value.
+  if [ $? -eq 0 ]; then
+
+    echo "info: parameter $parm_key found, updating its value"
+
+    sed -E "s/(^[[:blank:]]*)$parm_key=(.*$)/$key_value/g" \
+      $parm_file > $parm_file_updated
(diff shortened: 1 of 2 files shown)


mtschaffler:
Great extension - thanks 👍 !

bin/zis-parmlib-update.sh
```diff
@@ -0,0 +1,284 @@
...
+#   * The original value is returned
+#################################################
+resolve_env_parameter() {
+
+  parm=$1
+
+  # Are we dealing with an env variable based value?
+  echo $parm | grep -E "^[$]{1}[A-Z0-9_]+$" 1>/dev/null
+  # Yes, resolve the value using eval, otherwise return the value itself.
+  if [ $? -eq 0 ]; then
```
fkovinAtRocket: Simplification:
```
if echo $parm | grep -E -q "^[$]{1}[A-Z0-9_]+$"; then
  ...
```

bin/zis-parmlib-update.sh
```diff
@@ -0,0 +1,284 @@
...
+    ( eval "echo $parm" 2>/dev/null )
+    if [ $? -ne 0 ]; then
```
fkovinAtRocket: what is this supposed to do?

bin/zis-parmlib-update.sh
```diff
@@ -0,0 +1,284 @@
...
+  # Make sure the temporary file produced by sed is removed.
+  trap "rm $parm_file_updated" EXIT
+
+  echo "info: adding/updaing $parm_key with value $parm_value"
+
+  # Try to find the key in our parameter file.
+  grep "^[[:blank:]]*$parm_key=" $parm_file 1>/dev/null
+
+  # Does the key already exist? If yes, subtitute the value.
+  if [ $? -eq 0 ]; then
```
fkovinAtRocket: `if grep ...; then`

bin/zis-parmlib-update.sh
```diff
@@ -0,0 +1,284 @@
...
+    echo "error: value for $parm_key has not been provided"
+    exit 8
+  fi
+
+  key_value="$parm_key=$parm_value"
+  parm_file_updated="$parm_file"_updated
+
+  # Make a temp file for sed.
+  touch $parm_file_updated
+  if [ $? -ne 0 ]; then
```
fkovinAtRocket: why not set an ERR trap to catch this?

bin/zis-parmlib-update.sh
```diff
@@ -0,0 +1,284 @@
...
+  grep "^[[:blank:]]*$parm_key=" $parm_file 1>/dev/null
+
+  # Does the key already exist? If yes, subtitute the value.
+  if [ $? -eq 0 ]; then
+
+    echo "info: parameter $parm_key found, updating its value"
+
+    sed -E "s/(^[[:blank:]]*)$parm_key=(.*$)/$key_value/g" \
+      $parm_file > $parm_file_updated
+    if [ $? -ne 0 ]; then
```
fkovinAtRocket: why not set an ERR trap to catch this?

bin/zis-parmlib-update.sh
```diff
@@ -0,0 +1,284 @@
...
+      exit 8
+    fi
+
+    # Copy the updated file. Make sure the cp knows it's text and does the
+    # correct conversion.
+    cp -T -O u $parm_file_updated $parm_file
+    if [ $? -ne 0 ]; then
```
fkovinAtRocket: also ERR trap

bin/zis-parmlib-update.sh
ifakhrutdinov: My check seems more readable to me.

bin/zis-parmlib-update.sh
ifakhrutdinov: Yes, it will.

bin/zis-parmlib-update.sh
daveyc: I think Fyodor may have a good point here :)

ifakhrutdinov: 2:1, you win.

daveyc: IIRC, ```if``` is a command so always returns a 0 return code so it's a bug

bin/zis-parmlib-update.sh
```diff
@@ -0,0 +1,284 @@
...
+  # Input parameters.
+  parm_file=$1
+
+  echo "info: updating $ZIS_PARMLIB($ZIS_PARMLIB_MEMBER) with the values from\
+  $parm_file"
+
+  # Go over the provided SAMPLIB file, pick key-value pairs and update the ZIS
+  # PARMLIB member with those new values.
+  while IFS= read -r line; do
+    key_value_pattern="^[[:blank:]]*[[:alnum:].]*[[:blank:]]*=\
```
fkovinAtRocket: `[:alnum:].` - is the dot intentional?

bin/zis-parmlib-update.sh
ifakhrutdinov: No, this check should be correct. ```if``` executes the command first and checks the return code, it should not affect ```$?``` of ```grep```, I double-checked.

bin/zis-parmlib-update.sh
fkovinAtRocket: I tried: it did not. Why would it? `echo $parm` expands to just `echo` which just echos an empty string and exits.

ifakhrutdinov: Did you issue ```set -u```? It'll cause an error if an undefined variable is being dereferenced.

bin/zis-parmlib-update.sh
ifakhrutdinov: I haven't set up any ERR traps.

daveyc: My bad. We save the return code because we print it in the message and it gets clobbered by echo. Sorry!

fkovinAtRocket: Ah, ok. Though, I'm still not sure why we need two nested subshells: the parenthesis are one, and the `eval` is another. `(set -u && echo $foo) 2>/dev/null` seems to be enough

bin/zis-parmlib-update.sh
fkovinAtRocket: Ah, ok. Though, I'm still not sure why we need two nested subshells: the parenthesis are one, and the `eval` is another. `(set -u && echo $foo) 2>/dev/null` seems to be enough

fkovinAtRocket:
Some general notes:
- as much as I myself love Unix pipes and redirects and use them everywhere, I know that their behaviour in USS depends on the _TAG_REDIR.. and other variables. We need to keep that in mind. Do we ensure that the environment where they're run will have the variables set correctly?
- Something that I missed in the initial script: given that we have an installer, we'll also need an uninstaller. Maybe not now, but we'll need to think of it eventually

bin/zis-parmlib-update.sh
ifakhrutdinov: Let's say ```$parm='$FOO'```. ```echo``` wouldn't be able to resolve ```'$FOO'```, it would just print the name of the variable, i.e. you'd see ```$FOO```, not the value behind FOO.

daveyc:
@fkovinAtRocket IIRC, _TAG_REDIR is only relevant for enhanced ASCII which is not relevant to this PR.

ifakhrutdinov:
@fkovinAtRocket @daveyc thank you for the review! I will adjust the code according to your recommendations.

fkovinAtRocket:
@daveyc yeah, I guess, TAG_REDIR will become relevant only if you run it with ported Bash, which uses Enhanced ASCII, and actually enables it for pipes and redirects, among other things. 
The scripts say `#!/bin/sh` of course, so it can become a problem only if someone deliberately runs then with Bash instead, which is unlikely/unsupported.

bin/zis-parmlib-update.sh
ifakhrutdinov: Done.

bin/zis-parmlib-update.sh
ifakhrutdinov: Done.

bin/zis-parmlib-update.sh
ifakhrutdinov: Done.

bin/zis-parmlib-update.sh
ifakhrutdinov: Add an error trap and used an inline if check.

PR #98, opened 2019-10-29
Add the ability to skip SAF check in ZIS when the caller is privileged.

ifakhrutdinov:
The pull-request depends on the [corresponding zowe-common-c PR](https://github.com/zowe/zowe-common-c/pull/91).

This enhancement adds a new public function that allows calling ZIS without SAF checks if the caller is either:
* Supervisor state
* System key 0 - 7

Misc changes:
* Fix bugs in log messages in ZIS


ifakhrutdinov:
Regression tests have been successfully performed by @dmitrynikolaev, merging

PR #109, opened 2019-11-14
Fix ZIS debug mode activation

vit-tomica:
A fix to activate debug mode using DEBUG parameter keyword for the ZWESIS01 program.


vit-tomica:
Just FYI, I don't have privileges to merge the PR. I guess it's up to you then. Thanks.

PR #114, opened 2019-11-21
Bugfixes and features for v1.7.1

jordanfilteau1995:
Signed-off-by: Jordan Filteau <jfilteau@rocketsoftware.com>

### DIFF
--- a/c/zis/services/secmgmttss.c
+++ b/c/zis/services/secmgmttss.c
@@ -219,7 +219,7 @@ static unsigned int getAcidsInLine(const RadminVLText *currentLine, TSSAcidData 
   unsigned int acidCounter = 0;
   for (int i = 0; i < 4; i++) {
     char tmpBuffer[11] = {0}; /* Acid - 8, Admin Indicator - 3 */
-    
+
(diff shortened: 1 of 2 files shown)


jordanfilteau1995:
This fixes the following bugs:

- It would cut off the end of an acid when getting the access list of a group, resulting in invalid output in the response.

- All of the different administrator suffixes weren't defined, so it was incorrectly returning administrators.

This adds the following features:

- A backup routine for when a non-administrator tries to access the API. Instead of executing privileged commands and failing, it will execute a command to get their profile, and return only the information in their scope. This is a feature that most people won't need, since you'd ideally want to be an administrator if you were using this API, but the functionality is there.

- The ability to retrieve profiles only by prefix. This can be done by looking for a profile with a "." at the end. This will act as a wildcard which extracts everything matching that prefix.

jordanfilteau1995:
These fixes are important in maintaining consistency with the API when called from a system with RACF.

1000TurquoisePogs:
I believe this is a repeat of #112 but for RC, which is fine.

PR #119, opened 2019-12-02
load environment config to zss 

NakulManchanda:
Do - not - merge, created to discuss changes only yet
Signed-off-by: Nakul Manchanda <nakul.manchanda@ibm.com>

### DIFF
--- a/c/zss.c
+++ b/c/zss.c
@@ -343,6 +343,19 @@ TraceDefinition traceDefs[] = {
   {0,0}
 };
 
+static JsonObject *readEnvSettings(ShortLivedHeap *slh) {
+  char errorBuffer[512] = { 0 };
+  char envConfig[1024];
+  sprintf(envConfig, "{\n \"zssPort\": %s }", "12345");
+  Json *envSettings = jsonParseString(slh, envConfig, errorBuffer, sizeof(errorBuffer));
+  JsonObject *envSettingsJsonObject = NULL;
+    if (jsonIsObject(envSettings)) {
+      envSettingsJsonObject = jsonAsObject(envSettings);
+    }
+
+    return envSettingsJsonObject;
+}
+
 static JsonObject *readServerSettings(ShortLivedHeap *slh, const char *filename) {
 
   char jsonErrorBuffer[512] = { 0 };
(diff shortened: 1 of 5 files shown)


NakulManchanda:
added new `envService.c`, it has interface `readEnvSettings(prefix)` which loads all environment variable to zss, calling this in zss.c just before loading mvdSettings.

Todo:
Ideal next step is to override mvdSettings with overrides in envSettings
Also, extend envSetting to support json and array, right now it supports string or numbers

NakulManchanda:
[Extension in zss.c](https://github.com/NakulManchanda/zss/blob/b073d1c4d2f52f9db435831f4d29253b09e0f28d/c/zss.c#L1012)
[envService.c](https://github.com/NakulManchanda/zss/blob/b073d1c4d2f52f9db435831f4d29253b09e0f28d/c/envService.c#L109)

NakulManchanda:
// valid json key-value from environment variable
  export ZWED_AGENT_PORT=1234
  export ZWED_PLUGIN_PATH=/u/user
  export ZWED_BOOLEAN_TRUE_TEST=true
  export ZWED_BOOLEAN_FALSE_TEST=false
  export ZWED_ARRAY_TEST=[30,20]
  export ZWED_ARRAY_TEST2=[true,false]

// NOT SUPPORRTED: they both loose quotes
  export ZWED_ARRAY_TEST=["a","b"] // not supported array of strings
  export ZWED_JSON_TEST={"a":2,"c":"ddd"} //not supported json values

NakulManchanda:
added these four environment overrides
export ZWED_agent_http_port=
export ZWED_pluginsDir='
export ZWED_agent_http_ipAddresses=
export ZWED_privilegedServerName=

c/zss.c
```diff
@@ -731,18 +767,25 @@ static void printZISStatus(HttpServer *server) {
 
 }
 
-static void readAgentAddressAndPort(JsonObject *serverConfig, char **address, int *port) {
+static void readAgentAddressAndPort(JsonObject *serverConfig, JsonObject *envConfig, char **address, int *port) {
+  *port = jsonObjectGetNumber(envConfig, "ZWED_agent_http_port");
```
1000TurquoisePogs: This does not appear to be the same JSON as what app server builds. It would be simpler to just avoid the JSON idea entirely rather than have it incompatible:

This:
```
{
"ZWED_agent_http_port": 12345,
"ZWED_agent_http_ipAddresses": "1.2.3.4",
}
```

app server:

```
{
  "agent": {
    "http": {
      "port": 12345,
      "ipAddresses": "1.2.3.4"
    }
  }
}
```

The reason to have JSON at all was to be able to merge, so if not merging, no need for JSON.

1000TurquoisePogs:
Closing this due to stagings/cupids existing, which is based on this. It's looking to be working as intended.

PR #125, opened 2019-12-12
Ensure correct printing of ZIS parameters

### LINKED ISSUE #122: ZSS address space - S0C4  RSN=00000011
I'm not sure if this is the right place to report this but I am currently trying to get ZOWE running but right after the startup of the ZSS address space starts I see these messages and then it crashes:
```
MSGID     MSGTEXT
ZWES0001I ZSS Cross-Memory Server starting, version is 1.0.5
ZWES0002I Input parameters at 0x6FE6:
ZWES0101I 00000000  0016D5C1 D4C57EE9 E6C5E2C9 E26DE2E3 |..NAME=ZWESIS_ST|
```
The dump is a SYSTEM COMPLETION CODE=0C4  REASON CODE=00000011

ifakhrutdinov:
This fixes #122

### DIFF
--- a/c/zis/server.c
+++ b/c/zis/server.c
@@ -1166,7 +1166,7 @@ static int loadConfig(ZISContext *context,
             ZIS_LOG_INPUT_PARM_MSG, mainParms);
     zowedump(NULL, LOG_COMP_STCBASE, ZOWE_LOG_INFO,
              (void *)mainParms,
-             sizeof(mainParms) + mainParms->textLength);
+             sizeof(ZISMainFunctionParms) + mainParms->textLength);
   }
 
   context->parms = zisMakeParmSet();


c/zis/server.c
```diff
@@ -1166,7 +1166,7 @@ static int loadConfig(ZISContext *context,
             ZIS_LOG_INPUT_PARM_MSG, mainParms);
     zowedump(NULL, LOG_COMP_STCBASE, ZOWE_LOG_INFO,
              (void *)mainParms,
```
gadyck: As I read this, you are including the binary halfword length of the parms in the data that is being written, which does not seem correct to me.  I think what you want is-
zowedump(NULL, LOG_COMP_STCBASE, ZOWE_LOG_INFO,
                (void *)(mainParms+sizeof(mainParms->textLength),
                 mainParms->textLength);

ifakhrutdinov: @gadyck the initial intention was to dump the length halfword too.

gadyck: Okay.  It seemed odd, but I did not know exactly what zowedump did.

PR #137, opened 2019-12-31
chown support

sbollingerAtRocket:
Signed-off-by: Steven Bollinger <sbollinger@rocketsoftware.com>

### DIFF
--- a/c/unixFileService.c
+++ b/c/unixFileService.c
@@ -942,10 +942,6 @@ static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *respons
     setDefaultJSONRESTHeaders(response);
     addStringHeader(response, "Allow", "POST");
     writeHeader(response);
-
-    jsonStart(out);
-    jsonEnd(out);
-
     finishResponse(response);
   }
   return 0;
--- a/c/unixFileService.c
+++ b/c/unixFileService.c
@@ -920,6 +920,38 @@ static int serveUnixFileMetadata(HttpService *service, HttpResponse *response) {
   return 0;
 }
 
+static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *response) {
+  HttpRequest *request = response->request;
+  char *routeFileFrag = stringListPrint(request->parsedFile, 2, 1000, "/", 0);
+  char *encodedRouteFileName = stringConcatenate(response->slh, "/", routeFileFrag);
+  char *routeFileName = cleanURLParamValue(response->slh, encodedRouteFileName);
+
+  char *userId    = getQueryParam(response->request, "user");
+  char *groupId   = getQueryParam(response->request, "group");
+  char *pattern   = getQueryParam(response->request, "pattern");
+  char *recursive = getQueryParam(response->request, "recursive");
+
+  if (!strcmp(request->method, methodPOST)) {
+    directoryChangeOwnerAndRespond (response, routeFileName,
+                          userId, groupId, recursive, pattern);
+  }
+  else {
+    jsonPrinter *out = respondWithJsonPrinter(response);
+
+    setResponseStatus(response, 405, "Method Not Allowed");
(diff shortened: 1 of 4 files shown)


c/unixFileService.c
```diff
@@ -920,6 +920,38 @@ static int serveUnixFileMetadata(HttpService *service, HttpResponse *response) {
...
+  }
+  else {
+    jsonPrinter *out = respondWithJsonPrinter(response);
+
+    setResponseStatus(response, 405, "Method Not Allowed");
+    setDefaultJSONRESTHeaders(response);
+    addStringHeader(response, "Allow", "POST");
+    writeHeader(response);
+
+    jsonStart(out);
```
ifakhrutdinov: Do we really need this empty JSON? If not, we can get rid of the `respondWithJsonPrinter` call.

c/unixFileService.c
```diff
@@ -920,6 +920,38 @@ static int serveUnixFileMetadata(HttpService *service, HttpResponse *response) {
   return 0;
 }
 
+static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *response) {
+  HttpRequest *request = response->request;
+  char *routeFileFrag = stringListPrint(request->parsedFile, 2, 1000, "/", 0);
```
ifakhrutdinov: Can we check the input parameters? For example, `directoryChangeOwnerAndRespond` function uses `recursive` as is, without checking for NULL. Currently it's not clear which parameters are optional.

ifakhrutdinov:
Can you pleas also link the pull-requests this pull-request depends on? Otherwise it may get merged without the corresponding changes in zowe-common-c.

c/unixFileService.c
sbollingerAtRocket: The jira ticket specifies that we support the POST for this feature.  If you want to remove the message indicating what the is error in the request, then we can remove it.  Otherwise, I would keep it in.

c/unixFileService.c
sbollingerAtRocket: The code uses strcmp to set the bool recursive.  strcmp will return false if the string is null.  We can add an additional test for null to make it more clear.

c/unixFileService.c
ifakhrutdinov: I was talking about the lines:
```c
jsonStart(out);
jsonEnd(out);
```

c/unixFileService.c
ifakhrutdinov: > strcmp will return false if the string is null. 

It may work or it may segfault/S0C4, I think it's implementation defined.

daveyc: On z/OS strcmp() will just invoke a builtin which boils down to a CLST instruction with no check for NULL. I suspect it will result in a 0C4.

daveyc: FWIW, I don't use strcmp! I prefer to use macros or inline functions that check for null. 

```
inline bool streq(const char * lhs, const char * rhs) {            
  return lhs != NULL && rhs != NULL && strcmp(lhs, rhs) == 0;      
}                                                                  
```

sbollingerAtRocket: I have put in the additional test for NULL before the strcmp so there shouldn't be a problem.

c/unixFileService.c
sbollingerAtRocket: I can remove this is un-necessary.  Make sure it is;  there are 7 other places in this file with the exact code sequence.

ifakhrutdinov: You don't need to fix everything. If this is not necessary, why add more of it.

c/unixFileService.c
1000TurquoisePogs: I see this here now: https://github.com/zowe/zowe-common-c/pull/104/files#diff-5a6b6901a39402924b29b573b54bbc0eR571

PR #141, opened 2020-01-11
Return ASCRE RC and RSN in case of failure

ifakhrutdinov:


ifakhrutdinov:
The Jenkins status seems to be stuck here, but on the website it's "success", merging.

PR #144, opened 2020-01-16
Create Qsam data set

sbollingerAtRocket:
Signed-off-by: Steven Bollinger <sbollinger@rocketsoftware.com>

### DIFF
--- a/c/datasetService.c
+++ b/c/datasetService.c
@@ -18,7 +18,6 @@
 #include <metal/stdlib.h>
 #include <metal/string.h>
 #include "metalio.h"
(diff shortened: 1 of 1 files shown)


John-A-Davies:
As @ifakhrutdinov points out, QSAM is not a dataset type, it's Queued Sequential Access Method.  What we're providing is the creation of sequential datasets on DASD.  These datasets can be partitioned, or not.  To be fair, such datasets are commonly referred to as "QSAM datasets".  The main feature is that you access logical records sequentially, so no VSAM or ISAM datasets.

PR #152, opened 2020-02-03
Unify the use of the security management class

ifakhrutdinov:


c/securityService.c
```diff
@@ -1063,6 +1063,14 @@ static void respondToProfileGET(ClassMgmtCommonParms *commonParms,
 
   zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2, "%s begin\n", __FUNCTION__);
 
+  if (strlen(commonParms->className) > 0) {
+    respondWithError(response, HTTP_STATUS_FORBIDDEN,
+                     "non standard class not allowed");
```
rocketjared: In my opinion, this configuration detail (that Zowe security has been configured with a non-standard SAF Class name) should be logged, but Not returned to the browser client.

c/securityService.c
```diff
@@ -1063,6 +1063,14 @@ static void respondToProfileGET(ClassMgmtCommonParms *commonParms,
...
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+           "non standard class provided for profiles GET, leaving...\n");
+    return;
+  }
+
   if (strlen(commonParms->profileName) > 0) {
     respondWithError(response, HTTP_STATUS_NOT_IMPLEMENTED,
                      "specific profile info retrieval not implemented");
```
rocketjared: Recommend logging on server but not providing this detail to the HTTP client.

c/securityService.c
```diff
@@ -1391,6 +1399,14 @@ static void respondToProfileAccessListGET(ClassMgmtCommonParms *commonParms,
 
   zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2, "%s begin\n", __FUNCTION__);
 
+  if (strlen(commonParms->className) > 0) {
+    respondWithError(response, HTTP_STATUS_FORBIDDEN,
+                     "non standard class not allowed");
```
rocketjared: Log on server but don't provide detail to HTTP client.

c/securityService.c
ifakhrutdinov: @rocketjared by non-standard class the message means any class except for the `default-class` value passed via the REST request. No configuration details are exposed. Even if you pass the same explicit class name as the one in the config, you'll get this error.

rocketjared: As a general rule, web servers should avoid lacing 403 responses with details about why the request is being forbidden.  If details actionable by the system administrator are being placed in the Zowe log, what is the positive case for returning this detail to the client?

ifakhrutdinov: @rocketjared initially just for development. I agree, it's better to err on the side of caution. I've updated the PR.

PR #155, opened 2020-02-19
Add insightful port message

1000TurquoisePogs:

Before:
```
ZSS server settings: address=0.0.0.0, port=12224
Server startup problem ret=1115 reason=0x744c7247
```

After:
```
ZSS server settings: address=0.0.0.0, port=12224
Server startup problem ret=1115 reason=0x744c7247
This is usually due to the server port (12224) already being occupied. Is ZSS running twice?
```

### DIFF
--- a/c/zss.c
+++ b/c/zss.c
@@ -1107,7 +1107,10 @@ int main(int argc, char **argv){
       mainHttpLoop(server);
 
     } else{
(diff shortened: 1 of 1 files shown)


1000TurquoisePogs:
OK, I will do that.

PR #161, opened 2020-03-09
Password Reset

sakeerthy:
Goes along with:
https://github.com/zowe/zss/pull/161
https://github.com/zowe/zowe-common-c/pull/128
https://github.com/zowe/zlux-app-server/pull/106
https://github.com/zowe/zss-auth/pull/29
https://github.com/zowe/zlux-server-framework/pull/181
https://github.com/zowe/zlux-app-manager/pull/193

### DIFF
--- a/c/authService.c
+++ b/c/authService.c
@@ -38,6 +38,15 @@
 #define SAF_CLASS "ZOWE"
 #define JSON_ERROR_BUFFER_SIZE 1024
 
+#define PASSWORD_RESET_RC_OK                0
+#define PASSWORD_RESET_RC_WRONG_PASSWORD    111
+#define PASSWORD_RESET_RC_WRONG_USER        143
+#define PASSWORD_RESET_RC_TOO_MANY_ATTEMPTS 163
+#define PASSWORD_RESET_RC_NO_NEW_PASSSWORD  168
+#define PASSWORD_RESET_RC_WRONG_FORMAT      169
+
+#define RESPONSE_MESSAGE_LENGTH             100
+
 /*
  * A handler performing the SAF_AUTH check: checks if the user has the
  * specified access to the specified entity in the specified class
@@ -174,6 +183,19 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
   return 0;
 }
 
+void respondWithJsonStatus(HttpResponse *response, const char *status, int statusCode, const char *statusMessage) {
+    jsonPrinter *out = respondWithJsonPrinter(response);
+    setResponseStatus(response,statusCode,(char *)statusMessage);
+    setDefaultJSONRESTHeaders(response);
+    writeHeader(response);
+
+    jsonStart(out);
+    jsonAddString(out, "status", (char *)status);
+    jsonEnd(out);
+
+    finishResponse(response);
+}
+
 int resetPassword(HttpService *service, HttpResponse *response) {
   int returnCode = 0, reasonCode = 0;
   HttpRequest *request = response->request;
@@ -185,6 +207,11 @@ int resetPassword(HttpService *service, HttpResponse *response) {
     char errBuf[JSON_ERROR_BUFFER_SIZE];
     char responseString[RESPONSE_MESSAGE_LENGTH];
 
+    if (nativeBody == NULL) {
+      respondWithJsonStatus(response, "No body found", HTTP_STATUS_BAD_REQUEST, "Bad Request");
+      return HTTP_SERVICE_FAILED;
+    }
+
     Json *body = jsonParseUnterminatedString(request->slh, nativeBody, inLen, errBuf, JSON_ERROR_BUFFER_SIZE);
     JsonObject *inputMessage = jsonAsObject(body);
     Json *username = jsonObjectGetPropertyValue(inputMessage,"username");
@@ -193,14 +220,20 @@ int resetPassword(HttpService *service, HttpResponse *response) {
     int usernameLength = 0;
     int passwordLength = 0;
     int newPasswordLength = 0;
-    if(username != NULL) {
-      usernameLength = strlen(jsonAsString(username));
+    if (username != NULL) {
+      if (jsonAsString(username) != NULL) {
+        usernameLength = strlen(jsonAsString(username));
+      }
     }
-    if(password != NULL) {
-      passwordLength = strlen(jsonAsString(password));
+    if (password != NULL) {
+      if (jsonAsString(password) != NULL) {
+        passwordLength = strlen(jsonAsString(password));
+      }
     }
-    if(newPassword != NULL) {
-      newPasswordLength = strlen(jsonAsString(newPassword));
+    if (newPassword != NULL) {
+      if (jsonAsString(newPassword) != NULL) {
+        newPasswordLength = strlen(jsonAsString(newPassword));
+      }
     }
     if (usernameLength == 0) {
       respondWithJsonStatus(response, "No username provided",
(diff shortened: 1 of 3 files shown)


c/authService.c
```diff
@@ -173,6 +174,76 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+
+  if (!strcmp(request->method, methodPOST)) {
+    char *inPtr = request->contentBody;
+    char *nativeBody = copyStringToNative(request->slh, inPtr, strlen(inPtr));
+    int inLen = nativeBody == NULL ? 0 : strlen(nativeBody);
+    char errBuf[JSON_ERROR_BUFFER_SIZE];
+
+    Json *body = jsonParseUnterminatedString(request->slh, nativeBody, inLen, errBuf, JSON_ERROR_BUFFER_SIZE);
+    JsonObject *inputMessage = jsonAsObject(body);
+    Json *username = jsonObjectGetPropertyValue(inputMessage,"username");
```
ifakhrutdinov: This needs to be "sanitized". If anyone these fields aren't present you pass a null pointer to the BPX server.

c/authService.c
```diff
@@ -173,6 +174,76 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+    if (newPasswordLength == 0) {
+      setResponseStatus(response, HTTP_STATUS_BAD_REQUEST, "ERROR");
+      jsonAddString(p, "status", "No new password provided");
+    } else {
+      resetZosUserPassword(jsonAsString(username),  jsonAsString(password), jsonAsString(newPassword), &returnCode, &reasonCode);
+
+      switch (returnCode) {
+        case PASSWORD_RESET_OK:
+          setResponseStatus(response, HTTP_STATUS_OK, "OK");
+          jsonAddString(p, "status", "Password Successfully Reset");
```
ifakhrutdinov: I don't think you can start writing using a json printer before writing the headers.

c/authService.c
```diff
@@ -173,6 +174,76 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+          setResponseStatus(response,HTTP_STATUS_BAD_REQUEST,"ERROR");
+          jsonAddString(p, "status", "No new password for expired password provided");
+          break;
+        case PASSWORD_RESET_WRONG_FORMAT:
+          setResponseStatus(response,HTTP_STATUS_BAD_REQUEST,"ERROR");
+          jsonAddString(p, "status", "Incorrect format for new password");
+          break;
+        default:
+          setResponseStatus(response,HTTP_STATUS_BAD_REQUEST,"ERROR");
+          char *responseString;
```
ifakhrutdinov: You are writing to an arbitrary address!

c/authService.c
```diff
@@ -173,6 +174,202 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+      writeHeader(response);
+
+      jsonStart(p);
+      jsonAddString(p, "status", "No password provided");
+      jsonEnd(p);
+
+      finishResponse(response);
+      return HTTP_SERVICE_FAILED;
+	} else if (newPasswordLength == 0) {
+      setResponseStatus(response, HTTP_STATUS_BAD_REQUEST, "ERROR");
```
1000TurquoisePogs: lack of indentation pains me :cry:

sakeerthy: its indented for me on notepad++ but when its here its not indented not sure if its github adjusting width

c/authService.c
```diff
@@ -173,6 +174,202 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+
+          jsonStart(p);
+          jsonAddString(p, "status", "An incorrect password has been entered too many times and the account has been locked. Please contact your administrator for further instructions.");
+          jsonEnd(p);
+
+          finishResponse(response);
+          return HTTP_SERVICE_FAILED;
+        default:
+          setResponseStatus(response,HTTP_STATUS_BAD_REQUEST,"ERROR");
+          setContentType(response,"text/json");
```
1000TurquoisePogs: These headers are everywhere. You can greatly reduce duplication via this function: https://github.com/zowe/zowe-common-c/blob/master/c/httpserver.c#L4010

c/authService.c
```diff
@@ -173,6 +174,202 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  } else {
+    setResponseStatus(response,HTTP_STATUS_METHOD_NOT_FOUND,"ERROR");
+    setContentType(response,"text/json");
+    addStringHeader(response,"Server","jdmfws");
+    addStringHeader(response,"Transfer-Encoding","chunked");
+    addStringHeader(response, "Cache-control", "no-store");
+    addStringHeader(response, "Pragma", "no-cache");
+    writeHeader(response);
+    jsonStart(p);
+    jsonAddString(p, "status", "Method Not Allowed");
```
1000TurquoisePogs: Also did you know, if you are just doing a start-string-end json, you can do: https://github.com/zowe/zowe-common-c/blob/master/c/httpserver.c#L4019

c/authService.c
1000TurquoisePogs: Emacs shows:
» }·else·if·(newPasswordLength·==·0)·{$
······setResponseStatus(response,·HTTP_STATUS_BAD_REQUEST,·"ERROR");$

The conditionals use tab, the other lines use spaces. This is apparently confusing github.
I can't figure out how this was done, because only part of the new code has tabs.
There should never be tabs in source code. If you could please clean this up I bet it would show correctly on github.

ifakhrutdinov: @sakeerthy please don't use tabs, replace your new tabs with spaces (2 spaces per tab) and it should solve this indentation issues

c/zss.c
```diff
@@ -370,11 +370,11 @@ static JsonObject *readServerSettings(ShortLivedHeap *slh, const char *filename)
     if (logLevels) {
       TraceDefinition *traceDef = traceDefs;
       while (traceDef->name != 0) {
-	int traceLevel = jsonObjectGetNumber(logLevels, (char*) traceDef->name);
-	if (traceLevel > 0) {
-	  traceDef->function(traceLevel);
-	}
-	++traceDef;
+    int traceLevel = jsonObjectGetNumber(logLevels, (char*) traceDef->name);
```
ifakhrutdinov: I think you missed these indentations.

h/authService.h
```diff
@@ -26,7 +26,17 @@
 #include "httpserver.h"
 #include "dataservice.h"
 
+#define PASSWORD_RESET_OK                0
```
ifakhrutdinov: Are you going to use these outside the C file? If not, please move them and add some prefix to indicate whether this is an RC, RSN or something else.

h/authService.h
```diff
@@ -26,7 +26,17 @@
...
+#define PASSWORD_RESET_WRONG_PASSWORD    111
+#define PASSWORD_RESET_WRONG_USER        143
+#define PASSWORD_RESET_TOO_MANY_ATTEMPTS 163
+#define PASSWORD_RESET_NO_NEW_PASSSWORD  168
+#define PASSWORD_RESET_WRONG_FORMAT      169
+
+#define RESPONSE_MESSAGE_LENGTH          100
+
 int installAuthCheckService(HttpServer *server);
+int resetPassword(HttpService *service, HttpResponse *response);
```
ifakhrutdinov: I don't think you're using this outside authService.c, please make it a static function.

c/authService.c
```diff
@@ -173,6 +174,86 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  HttpRequest *request = response->request;
+  
+  if (!strcmp(request->method, methodPOST)) {
+    char *inPtr = request->contentBody;
+    char *nativeBody = copyStringToNative(request->slh, inPtr, strlen(inPtr));
+    int inLen = nativeBody == NULL ? 0 : strlen(nativeBody);
+    char errBuf[JSON_ERROR_BUFFER_SIZE];
+    char responseString[RESPONSE_MESSAGE_LENGTH];
+
+    Json *body = jsonParseUnterminatedString(request->slh, nativeBody, inLen, errBuf, JSON_ERROR_BUFFER_SIZE);
```
ifakhrutdinov: `body` can be NULL. Please check.

c/authService.c
```diff
@@ -173,6 +174,86 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
 }
 
+int resetPassword(HttpService *service, HttpResponse *response) {
+  int returnCode = 0, reasonCode = 0;
+  HttpRequest *request = response->request;
+  
+  if (!strcmp(request->method, methodPOST)) {
+    char *inPtr = request->contentBody;
+    char *nativeBody = copyStringToNative(request->slh, inPtr, strlen(inPtr));
+    int inLen = nativeBody == NULL ? 0 : strlen(nativeBody);
```
ifakhrutdinov: Can you do anything if `nativeBody` is NULL. Maybe you should return error right away and leave.

c/authService.c
```diff
@@ -173,6 +174,86 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+      return HTTP_SERVICE_FAILED;
+    } else if (passwordLength == 0) {
+      respondWithJsonStatus(response, "No password provided",
+                            HTTP_STATUS_BAD_REQUEST, "Bad Request");
+      return HTTP_SERVICE_FAILED;
+    } else if (newPasswordLength == 0) {
+      respondWithJsonStatus(response, "No new password provided",
+                            HTTP_STATUS_BAD_REQUEST, "Bad Request");
+      return HTTP_SERVICE_FAILED;
+    } else {
```
ifakhrutdinov: This is not a bug, just a note on the structure. Since every check results in leaving the function, you can make all these `if-else`s `if`s, and remove the last `else` which will make the code much more readable.

c/authService.c
```diff
@@ -173,6 +174,86 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+          respondWithJsonStatus(response, "The new password format is incorrect. Please try again.",
+                                HTTP_STATUS_BAD_REQUEST, "Bad Request");
+          return HTTP_SERVICE_FAILED;
+        case PASSWORD_RESET_TOO_MANY_ATTEMPTS:
+          respondWithJsonStatus(response,
+                                "An incorrect password has been entered too many times and the account has been locked. Please contact your administrator for further instructions.",
+                                HTTP_STATUS_TOO_MANY_REQUESTS, "Bad Request");
+          return HTTP_SERVICE_FAILED;
+        default:
+          sprintf(responseString, "Password reset FAILED with return code: %d reason code: %d\n", returnCode, reasonCode);
```
ifakhrutdinov: Can you please use `snprintf` to make sure no new changes will cause a buffer overflow?

c/zss.c
```diff
@@ -551,6 +551,16 @@ static void installLoginService(HttpServer *server) {
   registerHttpService(server, httpService);
 }
 
+static void installZosPasswordService(HttpServer *server) {
+  zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG2, "begin %s\n", __FUNCTION__);
```
ifakhrutdinov: Perhaps you need the `end %s\n` counterpart message.

h/authService.h
sakeerthy: Its called in zss.c

ifakhrutdinov: Sorry, I assumed you had install function defined in the authService.c similar to the rest of the installer functions.

c/authService.c
```diff
@@ -173,6 +183,116 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+      return HTTP_SERVICE_FAILED;
+    } else if (newPasswordLength == 0) {
+      respondWithJsonStatus(response, "No new password provided",
+                            HTTP_STATUS_BAD_REQUEST, "Bad Request");
+      return HTTP_SERVICE_FAILED;
+    } else {
+      resetZosUserPassword(jsonAsString(username),  jsonAsString(password), jsonAsString(newPassword), &returnCode, &reasonCode);
+
+      if (returnCode == PASSWORD_RESET_RC_OK) {
+        respondWithJsonStatus(response, "Password Successfully Reset", HTTP_STATUS_OK, "OK");
```
ifakhrutdinov: If you changed this after my comment about `if-else` then you must have misinterpreted it. I didn't mean removing the switch case. I meant making the passwordLength, newPasswordLength, etc checks just normal `if` check without else since they return immediately. The you could also remove the `else` on line 250 making it more readable.

Anyway, it's fine.

h/authService.h
```diff
@@ -27,6 +27,8 @@
 #include "dataservice.h"
 
 int installAuthCheckService(HttpServer *server);
+static int resetPassword(HttpService *service, HttpResponse *response);
```
ifakhrutdinov: You don't need to declare a static function in a header. Please remove this.

c/authService.c
```diff
@@ -33,8 +33,19 @@
 #include "socketmgmt.h"
 #include "zis/client.h"
 #include "httpserver.h"
+#include "zssLogging.h"
 
 #define SAF_CLASS "ZOWE"
+#define JSON_ERROR_BUFFER_SIZE 1024
+
+#define PASSWORD_RESET_RC_OK                0
+#define PASSWORD_RESET_RC_WRONG_PASSWORD    111
```
ifakhrutdinov: Where do these values come from? Can you clarify that in a comment or in the names?

c/authService.c
ifakhrutdinov: Will you check this?

ifakhrutdinov:
@sakeerthy can you please go over the comments and resolve the ones which are resolved and reply to the ones which aren't resolved?

1000TurquoisePogs:
Merged this because it seems like no work was left to do. But, let us know if follow-up work is needed.

PR #162, opened 2020-03-13
Pick up the zowe-comon-c deps, optimize some sources

ifakhrutdinov:
#### Overview
* There has been charsets optimizations done in zowe-common-c which this pull-request will pick up
* It's been decided that we should start optimizing out code. The json, charset and collections seem like a low hanging fruit with regards to how much we gain by turning on the compiler optimizations
* The changes have been regression-tested by @struga0258

### DIFF
--- a/build/build_zss.sh
+++ b/build/build_zss.sh
@@ -30,6 +30,27 @@ echo "Version: $major.$minor.$micro"
 echo "Date stamp: $date_stamp"
 
 c89 \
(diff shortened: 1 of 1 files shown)


ifakhrutdinov:
@1000TurquoisePogs, @struga0258  are you okay with merging this? It'd be good if we could start automated tests against this new build, just in case.

PR #163, opened 2020-03-13
(Ready for review) Refactoring ZSS logs

DivergentEuropeans:
**(DONE)** ${ZSS}/c/zssLogging.c \ no logs
**(DONE)** ${ZSS}/c/zss.c \ Only zowelogs here, need to add IDs - yes, it needs messages ID but I expect most of the naked printf or INFO level zowelog calls to become DEBUG level thus no message ids will be needed
**(DONE)** ${ZSS}/c/serviceUtils.c \ no logs
**(DONE)** ${ZSS}/c/authService.c \ no logs, a few commented out ones
**(DONE)** ${ZSS}/c/omvsService.c \ <add zowelog, add IDs> an error here, and a zowelog already exists
**(DONE)** ${ZSS}/c/unixFileService.c \ zowe logs only, need IDs
**(DONE)** ${ZSS}/c/datasetService.c \ zowe logs only, need IDs
**(DONE)** ${ZSS}/c/envService.c \ no logs, a few commented out ones
**(DONE)** ${ZSS}/c/zosDiscovery.c \ <add zowe log, add IDs> pre-existing zowelogs
**(DONE)** ${ZSS}/c/securityService.c \ <add zowe log, add IDs> filled with zowelogs (too long to examine quickly) probably just need to add IDs
**(DONE)** ${ZSS}/c/zis/client.c \ no logs here
**(DONE)** ${ZSS}/c/serverStatusService.c one zowe log here, needs ID


c/zosDiscovery.c
```diff
@@ -523,7 +524,7 @@ int zosDiscoveryServiceInstaller(DataService *dataService, HttpServer *server){
     CrossMemoryServerName *priviligedServerName = getConfiguredProperty(server, HTTP_SERVER_PRIVILEGED_SERVER_PROPERTY);
     state->model = makeZOSModel(priviligedServerName);
     dataService->extension = state;
-    printf("%s end\n", __FUNCTION__);
+    zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG, "%s end\n", __FUNCTION__);
```
1000TurquoisePogs: Why is the contents in zosDiscovery.c LOG_COMP_ID_MVD_SERVER?
We have one called LOG_COMP_DISCOVERY
This is related to discovery.c but is the zss API wrapper.
Is it OK to use discovery here? @ifakhrutdinov may have an opinion.

build/build_zss.sh
```diff
@@ -72,7 +72,6 @@ c89 \
   ${COMMON}/c/logging.c \
   ${COMMON}/c/nametoken.c \
   ${COMMON}/c/zos.c \
-  ${COMMON}/c/impersonation.c \
```
1000TurquoisePogs: really?

DivergentEuropeans: Yes, it's in the script twice

c/zosDiscovery.c
DivergentEuropeans: @1000TurquoisePogs I have no strong opinion, this was left from Suneeth's best judgement. Personally? It doesn't seem right to me to categorize this with DISCOVERY which is labeled as "Zowe Discovery". I think we should create a new component ID or use the one that handles generic server interaction (ID_MVD_SERVER)

h/zssLogging.h
```diff
@@ -13,13 +13,570 @@
...
+#define LOG_COMP_GROUP_REQ_USER_PP_MSG_TEXT  "group name required for user POST/PUT\n"
+#define LOG_COMP_GROUP_REQ_USER_PP           LOG_COMP_GROUP_REQ_USER_PP_MSG_ID" "LOG_COMP_GROUP_REQ_USER_PP_MSG_TEXT
+
+#ifndef LOG_COMP_ACCESS_REQ_USER_PP_MSG_ID
+#define LOG_COMP_ACCESS_REQ_USER_PP_MSG_ID LOG_MSG_PRFX"1425W"
+#endif
+#define LOG_COMP_ACCESS_REQ_USER_PP_MSG_TEXT "access type not provided for user POST/PUT, leaving...\n"
+#define LOG_COMP_ACCESS_REQ_USER_PP          LOG_COMP_ACCESS_REQ_USER_PP_MSG_ID" "LOG_COMP_ACCESS_REQ_USER_PP_MSG_TEXT
+
+#ifndef LOG_COMP_UNK_TYPE_UCCJ_MSG_ID
```
1000TurquoisePogs: Not a fan of needing to conditional define because of how many lines this is overall.
Also, not a fan of LOG_COMP doing double duty, where one is for a component id and another is for a log id. These are different things. Why not call these LOG_MSG_ID or something similar?

DivergentEuropeans: Hey @1000TurquoisePogs , I attempted to make the structure resemble something that's already preexisting via: https://github.com/zowe/zowe-common-c/blob/fbd8d446a84186df3a996c1f0ced58dcad7717c5/h/crossmemory.h#L35

However, we do have a handful of lines of code duplication that we can maybe cut down on? I was hesitant on doing this because they're across different repos...

ifakhrutdinov: What about
```
ZSS_xxxxx_MSG_ID
ZSS_xxxxx_MSG_TEXT
ZSS_xxxxx_MSG
```

c/zosDiscovery.c
ifakhrutdinov: That installer function doesn't seem to be directly related to the discovery code, just boiler plate code to establish communication with zss. I would follow the same log component logic as the other installer functions we have.

1000TurquoisePogs: That's fine, it will keep it simple.

h/zssLogging.h
DivergentEuropeans: So mass remove "LOG_COMP" and make it "ZSS"? @ifakhrutdinov @1000TurquoisePogs

ifakhrutdinov: @DivergentEuropeans yes, also explicitly indicate that a define is a message by having `MSG` somewhere in that `#define`'s name

DivergentEuropeans: I like the ZSS_LOG idea. I'll refactor based on that

h/zssLogging.h
```diff
@@ -13,13 +13,570 @@
...
+#define ZSS_LOG_DS_NAME_HLQ_MSG_ID            LOG_MSG_PRFX"1038I"
+#endif
+#define ZSS_LOG_DS_NAME_HLQ_MSG_TEXT          "l1=%s\n"
+#define ZSS_LOG_DS_NAME_HLQ_MSG               ZSS_LOG_DS_NAME_HLQ_MSG_ID" "ZSS_LOG_DS_NAME_HLQ_MSG_TEXT
+
+#ifndef ZSS_LOG_DS_SERVING_NAME_MSG_ID
+#define ZSS_LOG_DS_SERVING_NAME_MSG_ID        LOG_MSG_PRFX"1039I"
+#endif
+#define ZSS_LOG_DS_SERVING_NAME_MSG_TEXT      "Serving: %s\n"
+#define ZSS_LOG_DS_SERVING_NAME               ZSS_LOG_DS_SERVING_NAME_MSG_ID" "ZSS_LOG_DS_SERVING_NAME_MSG_TEXT
```
ifakhrutdinov: Some `#define`s have the _MSG suffix and some don't. For example, this one. Can you please add _MSG for consistency?

h/zssLogging.h
```diff
@@ -13,13 +13,570 @@
...
+#define ZSS_LOG_UNABLE_METADATA_MSG_ID       LOG_MSG_PRFX"1202W"
+#endif
+#define ZSS_LOG_UNABLE_METADATA_MSG_TEXT     "Could not get metadata for file. Ret: %d, Res: %d\n"
+#define ZSS_LOG_UNABLE_METADATA_MSG          ZSS_LOG_UNABLE_METADATA_MSG_ID" "ZSS_LOG_UNABLE_METADATA_MSG_TEXT
+
+#ifndef ZSS_LOG_UNABLE_OPEN_MSG_ID
+#define ZSS_LOG_UNABLE_OPEN_MSG_ID           LOG_MSG_PRFX"1203W"
+#endif
+#define ZSS_LOG_UNABLE_OPEN_MSG_TEXT         "Could not open file. Ret: %d, Res: %d\n"
+#define ZSS_LOG_UNABLE_OPEN_MSG              ZSS_LOG_UNABLE_OPEN_MSG_ID" "ZSS_LOG_UNABLE_OPEN_MSG_TEXT
```
ifakhrutdinov: This seems to be unused. Is this intentional?

c/unixFileService.c
```diff
@@ -367,7 +367,7 @@ static int checkIfFileIsBusy(HttpResponse *response, char *encodedFileName, Unix
                   &reasonCode);
 
   if (*file == NULL) {
-    zowelog(NULL, LOG_COMP_ID_UNIXFILE, ZOWE_LOG_WARNING, "Could not open file. Ret: %d, Res: %d\n",
+    zowelog(NULL, LOG_COMP_ID_UNIXFILE, ZOWE_LOG_WARNING, ZSS_LOG_UNABLE_CLOSE_MSG,
```
ifakhrutdinov: Should this be `ZSS_LOG_UNABLE_OPEN_MSG`?

h/zssLogging.h
```diff
@@ -13,13 +13,570 @@
...
+#ifndef ZSS_LOG_PROF_REQ_GET_MSG_ID
+#define ZSS_LOG_PROF_REQ_GET_MSG_ID          LOG_MSG_PRFX"1401W"
+#endif
+#define ZSS_LOG_PROF_REQ_GET_MSG_TEXT        "profile not provided for profiles GET, leaving...\n"
+#define ZSS_LOG_PROF_REQ_GET_MSG             ZSS_LOG_PROF_REQ_GET_MSG_ID" "ZSS_LOG_PROF_REQ_GET_MSG_TEXT
+
+#ifndef ZSS_LOG_NON_STRD_POST_MSG_ID
+#define ZSS_LOG_NON_STRD_POST_MSG_ID         LOG_MSG_PRFX"1402W"
+#endif
+#define ZSS_LOG_NON_STRD_POST_MSG_TEXT       "non standard class provided for profiles POST, leaving...\n"
```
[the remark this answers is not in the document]
DivergentEuropeans: It's up to you. I don't think there are plans, seeing as content-wise there wouldn't be more to support, just a greater quantity of messages. But I could consolidate if you think it's worth it
I just don't see the overall positive impact.

It might even be bad. Imagine "This thing A failed to load" and "This thing B failed to load" have the same message ID... that might get confusing

Let's imagine to a tech support person, the meaning of the message now depends on what the literal text says rather than the ID. Which is what we're trying to avoid, I believe.
Thoughts? @1000TurquoisePogs

c/zss.c
```diff
@@ -658,17 +658,17 @@ static WebPluginListElt* readWebPluginDefinitions(HttpServer *server, ShortLived
...
-                zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "plugin identifier or/and pluginLocation was not found in %s\n", path);
+                zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, ZSS_LOG_PLUGIN_IDLOC_NFOUND_MSG, path);
               }
             } else {
-              zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, "error in file %s: expected top level object\n", path);
+              zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_SEVERE, ZSS_LOG_FILE_EXPECTED_TOP_MSG, path);
             }
           } else {
-            zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, "error while parsing %s: %s\n", errorBuffer);
+            zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_SEVERE, ZSS_LOG_PARS_GENERIC_MSG, errorBuffer);
```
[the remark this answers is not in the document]
DivergentEuropeans: Great catch

h/zssLogging.h
```diff
@@ -13,13 +13,570 @@
...
+#ifndef ZSS_LOG_DS_DELETING_IF_MSG_ID
+#define ZSS_LOG_DS_DELETING_IF_MSG_ID        LOG_MSG_PRFX"1041I"
+#endif
+#define ZSS_LOG_DS_DELETING_IF_MSG_TEXT      "Deleting if exists: %s\n"
+#define ZSS_LOG_DS_DELETING_IF_MSG           ZSS_LOG_DS_DELETING_IF_MSG_ID" "ZSS_LOG_DS_DELETING_IF_MSG_TEXT
+
+#ifndef ZSS_LOG_INSTALL_DS_CONT_MSG_ID
+#define ZSS_LOG_INSTALL_DS_CONT_MSG_ID       LOG_MSG_PRFX"1042I"
+#endif
+#define ZSS_LOG_INSTALL_DS_CONT_MSG_TEXT     "Installing dataset contents service\n"
```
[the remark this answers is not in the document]
DivergentEuropeans: See above comment @ifakhrutdinov

h/zssLogging.h
```diff
@@ -13,13 +13,570 @@
...
+
+#ifndef ZSS_LOG_INSTALL_DS_MTADTA_MSG_ID
+#define ZSS_LOG_INSTALL_DS_MTADTA_MSG_ID     LOG_MSG_PRFX"1044I"
+#endif
+#define ZSS_LOG_INSTALL_DS_MTADTA_MSG_TEXT   "Installing dataset metadata service\n"
+#define ZSS_LOG_INSTALL_DS_MTADTA_MSG        ZSS_LOG_INSTALL_DS_MTADTA_MSG_ID" "ZSS_LOG_INSTALL_DS_MTADTA_MSG_TEXT
+
+/* MVD Server (Discovery) */
+
+#ifndef ZSS_LOG_DISC_THIRD_NGIVEN_MSG_ID
```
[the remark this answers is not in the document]
DivergentEuropeans: I'll make them into debug for simplicity's sake

h/zssLogging.h
```diff
@@ -13,13 +13,570 @@
...
+#ifndef ZSS_LOG_DS_SERVING_NAME_MSG_ID
+#define ZSS_LOG_DS_SERVING_NAME_MSG_ID       LOG_MSG_PRFX"1039I"
+#endif
+#define ZSS_LOG_DS_SERVING_NAME_MSG_TEXT     "Serving: %s\n"
+#define ZSS_LOG_DS_SERVING_NAME_MSG          ZSS_LOG_DS_SERVING_NAME_MSG_ID" "ZSS_LOG_DS_SERVING_NAME_MSG_TEXT
+
+#ifndef ZSS_LOG_DS_UPDATING_IF_MSG_ID
+#define ZSS_LOG_DS_UPDATING_IF_MSG_ID        LOG_MSG_PRFX"1040I"
+#endif
+#define ZSS_LOG_DS_UPDATING_IF_MSG_TEXT      "Updating if exists: %s\n"
```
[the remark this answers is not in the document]
DivergentEuropeans: Sure

h/zssLogging.h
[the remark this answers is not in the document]
ifakhrutdinov: What are the system action and operator response to these messages? If they are the same, i.e. "Request rejected" and "Correct [parm_name] and contact support if the problem hasn't been resolved" then I think it's worth having a single template.

I didn't mean arbitrary things which fall into the same template, I meant related things. For example, https://github.com/zowe/zowe-common-c/blob/master/h/crossmemory.h#L476

h/zssLogging.h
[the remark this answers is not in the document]
ifakhrutdinov: This actually seems like a good example of messages which could reuse a message ID. Would the system action and operator response be different?

h/zssLogging.h
```diff
@@ -52,529 +52,483 @@ bool isLogLevelValid(int level);
...
 #define ZSS_LOG_PORT_OCCUP_MSG               ZSS_LOG_PORT_OCCUP_MSG_ID" "ZSS_LOG_PORT_OCCUP_MSG_TEXT
 
 /* MVD Server (Datasets) */
 
 #ifndef ZSS_LOG_DS_NAME_HLQ_MSG_ID
-#define ZSS_LOG_DS_NAME_HLQ_MSG_ID           LOG_MSG_PRFX"1038I"
+#define ZSS_LOG_DS_NAME_HLQ_MSG_ID           ZSS_LOG_MSG_PRFX"1038I"
 #endif
-#define ZSS_LOG_DS_NAME_HLQ_MSG_TEXT         "l1=%s\n"
+#define ZSS_LOG_DS_NAME_HLQ_MSG_TEXT         "L1=%s\n"
```
ifakhrutdinov: @DivergentEuropeans tbh, this also seems like a debug message, what do you think?

(I haven't reviewed every single text)

DivergentEuropeans: I agree

1000TurquoisePogs:
Thanks Irek! I will review now too.

c/zss.c
```diff
@@ -463,16 +463,16 @@ static JsonObject *readPluginDefinition(ShortLivedHeap *slh, char *pluginIdentif
...
-        zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "plugin identifier was not found in %s\n", path);
+        zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, ZSS_LOG_PLUGIN_ID_NFOUND_MSG, path);
       }
     } else {
-      zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "error in file %s: expected top level object\n", path);
+      zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, ZSS_LOG_FILE_EXPECTED_TOP_MSG, path);
     }
   } else {
-    zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "error while parsing file %s: %s\n", path, errorBuffer);
+    zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, ZSS_LOG_PARS_FILE_MSG, path, errorBuffer);
```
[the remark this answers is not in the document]
DivergentEuropeans: Would the server crash here? @1000TurquoisePogs Otherwise, if there's no return, then I assume the server doesn't stop. Warning maybe?

c/zss.c
```diff
@@ -463,16 +463,16 @@ static JsonObject *readPluginDefinition(ShortLivedHeap *slh, char *pluginIdentif
...
-          zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "expected plugin identifier %s, got %s\n", pluginIdentifier, identifier);
+          zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, ZSS_LOG_EXPEC_PLUGIN_ID_MSG, pluginIdentifier, identifier);
         }
       } else {
-        zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "plugin identifier was not found in %s\n", path);
+        zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, ZSS_LOG_PLUGIN_ID_NFOUND_MSG, path);
       }
     } else {
-      zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "error in file %s: expected top level object\n", path);
+      zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, ZSS_LOG_FILE_EXPECTED_TOP_MSG, path);
```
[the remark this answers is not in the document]
DivergentEuropeans: Would the server crash here? @1000TurquoisePogs Otherwise, if there's no return, then I assume the server doesn't stop. Warning maybe?

c/zss.c
```diff
@@ -379,7 +379,7 @@ static JsonObject *readServerSettings(ShortLivedHeap *slh, const char *filename)
     }
     dumpJson(mvdSettings);
   } else {
-    zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, "error while parsing ZSS settings from file %s: %s\n", filename, jsonErrorBuffer);
+    zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, ZSS_LOG_PARS_ZSS_SETTING_MSG, filename, jsonErrorBuffer);
```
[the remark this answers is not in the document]
DivergentEuropeans: 💩

PR #168, opened 2020-04-02
Fix overwriting of environ variable due to strtok

timgerstel:
This pull request fixes the following bug:

- GET requests to /server/agent/environment missing environment variables

This was due to the use of strtok which replaces each matched delimiter with a null term, thus making the "environ" global var unusable after a single successful iteration.

### DIFF
--- a/c/envService.c
+++ b/c/envService.c
@@ -33,7 +33,7 @@ static bool startsWith(const char *pre, const char *str) {
     return lenstr < lenpre ? false : memcmp(pre, str, lenpre) == 0;
 }
 
-static void splitEnvKeyValue(char buf[], char* array[]) {
+static char* splitEnvKeyValue(char buf[], char* array[]) {
   char *bufCpy;
   if(buf != NULL){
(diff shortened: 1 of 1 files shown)


NakulManchanda:
@timgerstel as we discussed

looks like a reasonable change, as strtok has side effects, so when envService called twice, it messes it on second request as per my understanding.
may be one of the **test** we can add that is to call 
`JsonObject *envSettings=envVarsToObject("ZWED");` **this multiple times ( 2 or more)**

There are some tests at bottom of envService.c
Set environment variable like this:
```
  export ZWED_AGENT_PORT=1234
  export ZWED_PLUGIN_PATH=/u/user
  export ZWED_BOOLEAN_TRUE_TEST=true
  export ZWED_BOOLEAN_FALSE_TEST=false
  export ZWED_ARRAY_TEST=[30,20]
  export ZWED_ARRAY_TEST2=[true,false]
```

cc:  @struga0258

PR #173, opened 2020-04-10
Added Support for Changing Log Levels via REST API

jordanfilteau1995:
We used to use, "rasService.c", but we don't anymore. I have a dire need to change the logging levels of data services while zss is running, so I added support for that in the process.

https://github.com/zowe/zowe-common-c/pull/136

To test via Postman:

Retrieve current logging level:
GET {host}:{port}/ras/traceLevel?componentName={plugin_id}:{dataservice_name}

Set new logging level:
PUT {host}:{port}/ras/traceLevel?componentName={plugin_id}:{dataservice_name}&level={0-5}

### DIFF
--- a/build/build_zss.sh
+++ b/build/build_zss.sh
@@ -123,8 +123,9 @@ c89 \
   ${ZSS}/c/zosDiscovery.c \
   ${ZSS}/c/securityService.c \
   ${ZSS}/c/zis/client.c \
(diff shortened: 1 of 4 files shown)


jordanfilteau1995:
It appears as though my macro to remove trailing space has made this file hard to read. I can go back and put the spaces back if needed.

PR #177, opened 2020-04-29
Adding zssServer.sh, configure.sh, and start.sh in packaging

sakeerthy:

### DIFF
--- a/bin/configure.sh
+++ b/bin/configure.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+# This program and the accompanying materials are
+# made available under the terms of the Eclipse Public License v2.0 which accompanies
+# this distribution, and is available at https://www.eclipse.org/legal/epl-v20.html
+# 
+# SPDX-License-Identifier: EPL-2.0
+# 
+# Copyright Contributors to the Zowe Project.
+
+
+# Required variables on shell:
+# - ROOT_DIR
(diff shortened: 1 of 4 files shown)


bin/zssServer.sh
```diff
@@ -0,0 +1,163 @@
...
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Copyright Contributors to the Zowe Project.
+
+## launch the Zowe Secure Services Server
+
+ZSS_SCRIPT_DIR=$(cd `dirname $0` && pwd)
+
+echo ${ZSS_SCRIPT_DIR}
```
1000TurquoisePogs, line 14: remove, not needed

bin/configure.sh
```diff
@@ -0,0 +1,29 @@
...
+if [ -n "$NODE_HOME" ]
+then
+  NODE_BIN=${NODE_HOME}/bin/node
+else
+  NODE_BIN=node
+fi
+
+cd ${ROOT_DIR}/components/app-server/share/zlux-app-server/bin
+. ./convert-env.sh
+
```
1000TurquoisePogs, line 24: Need to confirm if the nodejs env vars such as _TAG_REDIR_OUT=txt is set at this time, because otherwise it would need to be set explicitly here.

PR #182, opened 2020-05-13
Relative plugins

1000TurquoisePogs:
Small enhancement to allow plugin locations to be relative to some location other than the server
This is the same as https://github.com/zowe/zlux-server-framework/pull/212


1000TurquoisePogs:
James says this is OK to merge after testing, so going for it.

PR #183, opened 2020-05-13
Small bugfix for compile warnings

1000TurquoisePogs:


c/envService.c
```diff
@@ -37,7 +37,7 @@ static char* splitEnvKeyValue(char buf[], char* array[]) {
   char *bufCpy;
   if(buf != NULL){
     int bufferLen = strlen(buf);
-    bufCpy = safeMalloc(bufferLen + 1, "buffer copy");
+    bufCpy = (char*) safeMalloc(bufferLen + 1, "buffer copy");
```
ifakhrutdinov, line 40: The right way is to include `alloc.h`. When no declaration is provided, the compiler assumes that the return value and arguments are all `int`. When compiled under LP64, this will lead to an S0C4.

1000TurquoisePogs: thank you, that makes sense now that you say it.

PR #188, opened 2020-05-29
 Make ZSS build fail when RC != 0

ifakhrutdinov:
**Overview**

Currently, if a warning message is produced by the compiler, the build process is considered successful. This is often dangerous as warnings may mean passing the wrong type or redefinition of a `#define`, which should be considered bugs. The pull-requests makes the build process stricter - any warning messages will now lead to a failing build.

**Changes**
* Make sure there are no warning messages in the current build
  * Update deps to remove the httpfileservice.c warning message, and pick up a minor type fix
  * Ensure side-deck file/SYSDEFSD DD by adding the dll option to the linker
* Adjust the compiler env variable controlling the severity
* Ensure no ZSS binary is created if RC != 0

### DIFF
--- a/build/build_zss.sh
+++ b/build/build_zss.sh
@@ -29,6 +29,8 @@ date_stamp=$(date +%Y%m%d)
 echo "Version: $major.$minor.$micro"
 echo "Date stamp: $date_stamp"
 
+export _C89_ACCEPTABLE_RC=0
+
 c89 \
   -c -O2 \
   -DPRODUCT_MAJOR_VERSION="$major" \
@@ -62,7 +64,7 @@ c89 \
   -DAPF_AUTHORIZED=0 \
   -Wc,dll,expo,langlvl\(extc99\),gonum,goff,hgpr,roconst,ASM,asmlib\('CEE.SCEEMAC','SYS1.MACLIB','SYS1.MODGEN'\) \
   -Wc,agg,exp,list\(\),so\(\),off,xref \
-  -Wl,ac=1 \
+  -Wl,ac=1,dll \
   -I ${COMMON}/h \
   -I ${COMMON}/jwt/jwt \
   -I ${COMMON}/jwt/rscrypto \
--- a/build/build_zss.sh
+++ b/build/build_zss.sh
@@ -22,6 +22,8 @@ COMMON="../../deps/zowe-common-c"
 echo "********************************************************************************"
 echo "Building ZSS..."
 
+rm -f ${ZSS}/bin/zssServer
+
 mkdir -p "${WORKING_DIR}/tmp-zss" && cd "$_"
 
 IFS='.' read -r major minor micro < "${ZSS}/version.txt"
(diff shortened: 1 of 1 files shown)


build/build_zss.sh
```diff
@@ -62,7 +70,7 @@ c89 \
   -DAPF_AUTHORIZED=0 \
   -Wc,dll,expo,langlvl\(extc99\),gonum,goff,hgpr,roconst,ASM,asmlib\('CEE.SCEEMAC','SYS1.MACLIB','SYS1.MODGEN'\) \
   -Wc,agg,exp,list\(\),so\(\),off,xref \
-  -Wl,ac=1 \
+  -Wl,ac=1,dll \
```
1000TurquoisePogs, line 73: can you give a short description about the difference here? I see it is a supported flag but I am curious about what was wrong when we only had dll for -Wc, but not -Wl

build/build_zss.sh
```diff
@@ -22,14 +22,18 @@ COMMON="../../deps/zowe-common-c"
...
+
 mkdir -p "${WORKING_DIR}/tmp-zss" && cd "$_"
 
 IFS='.' read -r major minor micro < "${ZSS}/version.txt"
 date_stamp=$(date +%Y%m%d)
 echo "Version: $major.$minor.$micro"
 echo "Date stamp: $date_stamp"
 
-c89 \
+export _C89_ACCEPTABLE_RC=0
```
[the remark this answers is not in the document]
1000TurquoisePogs: ok, so it's not for the script, its for xlc. makes sense.

build/build_zss.sh
ifakhrutdinov: According to the [manual](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.1.0/com.ibm.zos.v2r1.cbcux01/oebind3.htm), the dll compiler options tell the compiler that some symbols will come from a dll, the dll linker option tells the linker to produce a side file and a proper dll.

I've just tested and the build works fine without the compiler dll option. However, the build will produce a warning if there is no linker dll option. I think it's because, we still have the export all option, which makes c89 create a binary with exported symbols, but a missing dll option causes the linker not to allocate the SYSDEFSD (side-deck DD) and produce a warning message.
```
When you bind code with exported symbols, you should specify the DLL
binder option (-W l,dll).
```

1000TurquoisePogs: One of the things this does is generate the side file with every single function. We don't really want that since it is a form of API documentation and we only want a subset of those functions to be externally used in case we change something. So, the .x file is going to be discarded anyway. This PR was an attempt to make an official one: https://github.com/zowe/zss/pull/30
So, is this still better even if we do not use the .x file that gets generated?

ifakhrutdinov: > So, is this still better even if we do not use the .x file that gets generated?
This is just to make the linker stop producing that warning message. That PR is still relevant. We need to select a subset of the functions we want to export. That is, right now it's "export all" later it'll be specific pragmas, but we'll have to use `-Wl,dll` in either case.

daveyc: That's acceptable to me. I don't know of any way to stop generating a side file with -qdll

PR #196, opened 2020-07-15
Change tag Recursively.

Akayeshmantha:


Akayeshmantha:
@1000TurquoisePogs  thanks for rebasing

c/unixFileService.c
```diff
@@ -981,6 +981,39 @@ static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *respons
   return 0;
 }
 
+static int serveUnixFileChangeTag (HttpService *service, HttpResponse *response) {
+  HttpRequest *request = response->request;
+  char *routeFileFrag = stringListPrint(request->parsedFile, 2, 1000, "/", 0);
+  char *encodedRouteFileName = stringConcatenate(response->slh, "/", routeFileFrag);
+  char *routeFileName = cleanURLParamValue(response->slh, encodedRouteFileName);
+
+  char *codepage  = getQueryParam(response->request, "codeset");
```
ifakhrutdinov, line 990: Any of these can return NULL.

c/unixFileService.c
```diff
@@ -981,6 +981,39 @@ static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *respons
...
+  char *recursive = getQueryParam(response->request, "recursive");
+  char *pattern   = getQueryParam(response->request, "pattern");
+  char *type      = getQueryParam(response->request, "type");
+
+  if (!strcmp(request->method, methodPOST)) {
+    directoryChangeTagAndRespond (response, routeFileName,
```
ifakhrutdinov, line 996: Can you please be consistent with formatting. No need to place a blank between a function name and the opening bracket.

c/unixFileService.c
```diff
@@ -981,6 +981,39 @@ static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *respons
...
+                                  type, codepage, recursive, pattern);
+  }
+  else if (!strcmp(request->method, methodDELETE)) {
+      char type[10];
```
ifakhrutdinov, line 1000: The following should simplify you logic.
```c
char type[8] = {"delete"};
```

c/unixFileService.c
```diff
@@ -981,6 +981,39 @@ static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *respons
...
+      strncpy (type, "delete", sizeof(type)-1);
+      directoryChangeDeleteTagAndRespond (response, routeFileName,
+                                  type, codepage, recursive, pattern);
+    }
```
ifakhrutdinov, line 1004: Bad indentation.

c/zss.c
```diff
@@ -1123,7 +1123,10 @@ static void initializePluginIDHashTable(HttpServer *server) {
 #define ZSS_STATUS_ERROR  8
 
 int main(int argc, char **argv){
-
+  if (argc == 1) {
```
ifakhrutdinov, line 1126: Why do we need an additional check here? Isn't line 1139 doing the same?

c/zss.c
```diff
@@ -1215,6 +1218,9 @@ int main(int argc, char **argv){
       installUnixFileTouchService(server);
       installUnixFileMetadataService(server);
       installUnixFileChangeOwnerService(server);
+#ifdef __ZOWE_OS_ZOS
```
ifakhrutdinov, line 1221: Why only ZOS?

c/unixFileService.c
```diff
@@ -981,6 +981,39 @@ static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *respons
...
+  else {
+    jsonPrinter *out = respondWithJsonPrinter(response);
+
+    setResponseStatus(response, 405, "Method Not Allowed");
```
ifakhrutdinov, line 1008: Can you use `respondWithError` instead of all of these lines?

c/unixFileService.c
```diff
@@ -981,6 +981,39 @@ static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *respons
   return 0;
 }
 
+static int serveUnixFileChangeTag (HttpService *service, HttpResponse *response) {
+  HttpRequest *request = response->request;
+  char *routeFileFrag = stringListPrint(request->parsedFile, 2, 1000, "/", 0);
```
ifakhrutdinov, line 986: Would you check these for NULL? I believe these function may fail.

c/zss.c
1000TurquoisePogs: `chtag` and tagging is an ability that only exists on ZOS.

c/unixFileService.c
```diff
@@ -981,6 +981,43 @@ static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *respons
...
+
+  if (routeFileFrag == NULL) {
+   respondWithError(response, HTTP_STATUS_BAD_REQUEST, "File name not provided");
```
ifakhrutdinov, line 989: Is this the right message? Shouldn't this be "Allocation failed" with `HTTP_STATUS_INTERNAL_SERVER_ERROR`? Same true for the next 2 calls.

c/unixFileService.c
```diff
@@ -981,6 +981,43 @@ static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *respons
...
+  char *pattern   = getQueryParam(response->request, "pattern");
+  char *type      = getQueryParam(response->request, "type");
+
+  if (codepage == NULL || recursive == NULL || pattern == NULL || type == NULL ) {
+   respondWithError(response, HTTP_STATUS_BAD_REQUEST, "Required parameter not provided");
+   return 0;
+  }
+
+  if (!strcmp(request->method, methodPOST)) {
+    directoryChangeTagAndRespond(response, routeFileName,type, codepage, recursive, pattern);
```
ifakhrutdinov, line 1007: I think you formatting is a bit arbitrary. Can you please use blanks after commas.

c/zss.c
ifakhrutdinov: Please see my previous comment.

c/unixFileService.c
Akayeshmantha: isn't it a bad request with the required params. I am not sure about this normally if a required param is not there it should be a bad request

c/unixFileService.c
Akayeshmantha: sure

c/zss.c
Akayeshmantha: yes it should be

c/unixFileService.c
1000TurquoisePogs: This sort of logic is used in a few places for checking if the url has the right shape 
https://github.com/zowe/zss/blob/master/c/rasService.c#L123
Seems like this can be the case when the stringlistprint returns with pos=0
https://github.com/zowe/zowe-common-c/blob/master/c/utils.c#L1681

also anyone know what this is https://github.com/zowe/zowe-common-c/blob/master/c/utils.c#L1658
seems like a weird thing to do.

ifakhrutdinov: From what I can see in `stringListPrint`, it would only return NULL if `SLHAlloc` failed. However, even if it fails we will segfault because the `stringListPrint` code doesn't check the `SLHAlloc` return value. In case at some point in the future we add a NULL check inside `stringListPrint`, the callers of this function will also need to check the return value for NULL. So it'd be reasonable to start doing that now. If we want to fix all the places at ones in the future, that's ok, but the messages is still misleading.

Akayeshmantha: @ifakhrutdinov can we meege this PR

c/unixFileService.c
```diff
@@ -981,6 +981,43 @@ static int serveUnixFileChangeOwner (HttpService *service, HttpResponse *respons
...
+
+  char *encodedRouteFileName = stringConcatenate(response->slh, "/", routeFileFrag);
+  char *routeFileName = cleanURLParamValue(response->slh, encodedRouteFileName);
+
+  char *codepage  = getQueryParam(response->request, "codeset");
+  char *recursive = getQueryParam(response->request, "recursive");
+  char *pattern   = getQueryParam(response->request, "pattern");
+  char *type      = getQueryParam(response->request, "type");
+
+  if (codepage == NULL || recursive == NULL || type == NULL) {
```
Akayeshmantha, line 1001: @ifakhrutdinov  I removed the null check for pattern because it is not a required parameter. And already there is a null check on the path

Akayeshmantha: aslo codepage if it is binary in type we can't define a code no it seems.

ifakhrutdinov: Do we want to make it required when it's not binary? What do similar services do?

c/unixFileService.c
ifakhrutdinov: @Akayeshmantha 
The NULL check at the top has the wrong message and passing an empty path isn't really handled.

image

c/unixFileService.c
Akayeshmantha: Yes we can make it required when the type is not binary

c/unixFileService.c
Akayeshmantha: Yes @ifakhrutdinov 😃.It is not handled my bad 😞

c/unixFileService.c
Akayeshmantha: its already been covered @ifakhrutdinov 

Screenshot from 2020-08-03 12-37-57

c/unixFileService.c
Akayeshmantha: edge cases when absolute path not provided.
Screenshot from 2020-08-03 20-10-52

Akayeshmantha: edge cases when absolute path is not provided.

Screenshot from 2020-08-03 21-51-40

Akayeshmantha: @ifakhrutdinov  please have a look

ifakhrutdinov: The NULL check is not correct since the result can only be NULL when the SLH fails which has nothing to do with this particular service. Please either remove it or use the following approach:

```c
  char *routeFileFrag = stringListPrint(request->parsedFile, 2, 1000, "/", 0);
  if (routeFileFrag == NULL) {
   respondWithError(response, "Allocation error", HTTP_STATUS_INTERNAL_SERVER_ERROR);
   return 0;
  }

  if (strlen(routeFileFrag) == 0) {
   respondWithJsonError(response, "Required absolute path of the resource is not provided", 
                        HTTP_STATUS_BAD_REQUEST, "Bad Request");
   return 0;
  }

  char *encodedRouteFileName = stringConcatenate(response->slh, "/", routeFileFrag);
  if (encodedRouteFileName == NULL) {
   respondWithError(response, "Allocation error", HTTP_STATUS_INTERNAL_SERVER_ERROR);
   return 0;
  }
  char *routeFileName = cleanURLParamValue(response->slh, encodedRouteFileName);
  if (routeFileName == NULL) {
   respondWithError(response, "Allocation error", HTTP_STATUS_INTERNAL_SERVER_ERROR);
   return 0;
  }
```

Akayeshmantha: sure @ifakhrutdinov . Yeah there is no null case

PR #204, opened 2020-08-05
Add the functionality to download folder content.

Akayeshmantha:


1000TurquoisePogs (changes requested):
You have conflicts that need resolving, but when I tried to resolve them and use your other pull requests, I got a failure to build:
```
./../c/rasService.c:
 IEW2456E 9207 SYMBOL createFileFromUnixDirectoryAndRespond UNRESOLVED.  MEMBER
          COULD NOT BE INCLUDED FROM THE DESIGNATED CALL LIBRARY.
```
I think you may have renamed a function without updating both branches.
It would be a good idea to update everything and add meaningful descriptions about how they are related and what they are supposed to do.

PR #205, opened 2020-08-05
Pass service param to use on force encoding.

Akayeshmantha:


ifakhrutdinov:
@Akayeshmantha can you clarify the end goal of the change? That is, what are you trying to achieve?

ifakhrutdinov (commented):
Please clarify the intent.

Akayeshmantha:
@ifakhrutdinov for the changes done in here  https://github.com/zowe/zowe-common-c/pull/160 in commons. We need to pass the service object from zss in order to capture the request params.

ifakhrutdinov:
The only place where the service parm is used in `respondWithUnixFile2` is this
```c
    if ((NULL != service) &&
        (NULL != service->customHeadersFunction))
    {
      service->customHeadersFunction(service, response);
    }
```
Can you point me to the place where you wanted to use it?

1000TurquoisePogs:
This needs a real description. Even if I know what it is right now, others don't and it's likely that a year from now if I need to review history, I won't remember.
I believe this is for the file transfer app feature where you may want to explicitly download as a particular encoding in the case that something is tagged incorrectly.

PR #208, opened 2020-08-18
Register ZSS to ML using configuration script

DivergentEuropeans:
This is similar to how the ML team registers their APIs (for example, like Jobs API)

### DIFF
--- a/bin/configure.sh
+++ b/bin/configure.sh
@@ -33,75 +33,27 @@ cat <<EOF >${STATIC_DEF_CONFIG_DIR}/zss.ebcidic.yml
 #
 services:
   - serviceId: zss
-    title: Zowe System Services Server (ZSS)
-    description: Zowe System Services Server (ZSS) is used for enabling low-level microservices and other privileged services
+    title: Zowe System Services (ZSS)
+    description: Zowe System Services is used for enabling low-level microservices and other privileged mainframe services, 
+    like USS, MVS, authentication, and security management.
     catalogUiTileId: zss
     instanceBaseUrls:
       - https://${ZOWE_EXPLORER_HOST}:${ZOWE_ZSS_SERVER_PORT}/
     homePageRelativeUrl:
     routedServices:
-      - gatewayUrl: /server/agent/environment # Displays ZSS host environment info
-        serviceRelativeUrl: /server/agent/environment 
-      - gatewayUrl: /server/agent/services  # Returns list of ZSS services
-        serviceRelativeUrl: /server/agent/services
-      - gatewayUrl: /plugins
-        serviceRelativeUrl: /plugins
-      - gatewayUrl: /datasetMetadata/
-        serviceRelativeUrl: /datasetMetadata/
-      - gatewayUrl: /datasetContents/
-        serviceRelativeUrl: /datasetContents/
-      - gatewayUrl: /security-mgmt/classes/
-        serviceRelativeUrl: /security-mgmt/classes/
-      - gatewayUrl: /security-mgmt/access/
-        serviceRelativeUrl: /security-mgmt/access/
-      - gatewayUrl: /security-mgmt/groups/
-        serviceRelativeUrl: /security-mgmt/groups/
-      - gatewayUrl: /security-mgmt/user-profiles/
-        serviceRelativeUrl: /security-mgmt/user-profiles/
-      - gatewayUrl: /saf-auth/
-        serviceRelativeUrl: /saf-auth/
-      - gatewayUrl: /VSAMdatasetContents/
-        serviceRelativeUrl: /VSAMdatasetContents/
-      - gatewayUrl: /unixfile/contents
-        serviceRelativeUrl: /unixfile/contents
-      - gatewayUrl: /unixfile/chmod/
-        serviceRelativeUrl: /unixfile/chmod/
-      - gatewayUrl: /unixfile/chtag/
-        serviceRelativeUrl: /unixfile/chtag/
-      - gatewayUrl: /unixfile/chown/
-        serviceRelativeUrl: /unixfile/chown/
-      - gatewayUrl: /unixfile/metadata/
-        serviceRelativeUrl: /unixfile/metadata/
-      - gatewayUrl: /unixfile/touch/
-        serviceRelativeUrl: /unixfile/touch/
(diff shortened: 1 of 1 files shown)


bin/configure.sh
```diff
@@ -27,3 +27,82 @@ export NODE_PATH=../..:../../zlux-server-framework/node_modules:$NODE_PATH
...
+services:
+  - serviceId: zss
+    title: Zowe System Services Server (ZSS)
+    description: Zowe System Services Server (ZSS) is used for enabling low-level microservices and other privileged services
+    catalogUiTileId: zss
+    instanceBaseUrls:
+      - https://${ZOWE_EXPLORER_HOST}:${ZOWE_ZSS_SERVER_PORT}/
+    homePageRelativeUrl:
+    routedServices:
+      - gatewayUrl: /server/agent/environment # Displays ZSS host environment info
```
lchudinov, line 43: All `gatewayUrl` should be with a prefix e.g. `/zss/api/v1/server/agent/environment`

lchudinov (changes requested):

bin/configure.sh
```diff
@@ -27,3 +27,82 @@ export NODE_PATH=../..:../../zlux-server-framework/node_modules:$NODE_PATH
...
+# Add static definition for zss. TODO: Needs routedServices + apiInfo
+cat <<EOF >${STATIC_DEF_CONFIG_DIR}/zss.ebcidic.yml
+#
+services:
+  - serviceId: zss
+    title: Zowe System Services Server (ZSS)
+    description: Zowe System Services Server (ZSS) is used for enabling low-level microservices and other privileged services
+    catalogUiTileId: zss
+    instanceBaseUrls:
+      - https://${ZOWE_EXPLORER_HOST}:${ZOWE_ZSS_SERVER_PORT}/
```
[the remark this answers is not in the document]
1000TurquoisePogs: It is confusing but that is actually the variable currently used for the hostname of zowe.

bin/configure.sh
1000TurquoisePogs: It doesnt seem that you need to prefix the name. Here is what I see on others
```
> cat datasets_ui.yml
#
services:
  - serviceId: explorer-mvs
    title: IBM z/OS Datasets UI
    description: IBM z/OS Datasets UI service
    catalogUiTileId:
    instanceBaseUrls:
      - https://hardcodedurl/
    homePageRelativeUrl:
    routedServices:
      - gatewayUrl: ui/v1
        serviceRelativeUrl: /ui/v1/explorer-mvs

#
services:
  - serviceId: explorer-jes
    title: IBM z/OS Jobs UI
    description: IBM z/OS Jobs UI service
    catalogUiTileId:
    instanceBaseUrls:
      - https://hardcodedurl/
    homePageRelativeUrl:
    routedServices:
      - gatewayUrl: ui/v1
        serviceRelativeUrl: /ui/v1/explorer-jes
```

bin/configure.sh
```diff
@@ -27,3 +27,82 @@ export NODE_PATH=../..:../../zlux-server-framework/node_modules:$NODE_PATH
...
+      - gatewayUrl: /logout/
+        serviceRelativeUrl: /logout/
+      - gatewayUrl: /ras/
+        serviceRelativeUrl: /ras/
+      - gatewayUrl: /password/
+        serviceRelativeUrl: /password/
+      - gatewayUrl: /omvs/
+        serviceRelativeUrl: /omvs/
+      - gatewayUrl: /unixfile/
+        serviceRelativeUrl: /unixfile/
```
1000TurquoisePogs, line 94: why do we have a zillion routes?
this just wont work for static because there would be no way to register plugin routes which I don't even see on here.

You could just have 1 route
gatewayUrl: /
serviceRelativeUrl: /
why not?

bin/configure.sh
```diff
@@ -27,3 +27,82 @@ export NODE_PATH=../..:../../zlux-server-framework/node_modules:$NODE_PATH
...
+    apiInfo:
+      - apiId: org.zowe.zss
+        gatewayUrl: server/agent/
```
1000TurquoisePogs, line 97: this is a bit misleading. you could omit it until we have something better.

1000TurquoisePogs (changes requested):

bin/configure.sh
```diff
@@ -27,3 +27,82 @@ export NODE_PATH=../..:../../zlux-server-framework/node_modules:$NODE_PATH
 __UNTAGGED_READ_MODE=V6 $NODE_BIN initInstance.js
 
 APP_WORKSPACE_DIR=${INSTANCE_DIR}/workspace/app-server
+
+# Add static definition for zss. TODO: Needs routedServices + apiInfo
+cat <<EOF >${STATIC_DEF_CONFIG_DIR}/zss.ebcidic.yml
+#
+services:
+  - serviceId: zss
+    title: Zowe System Services Server (ZSS)
```
1000TurquoisePogs, line 36: That would be ZSSS
You could just have it as
Zowe System Services (ZSS)

bin/configure.sh
1000TurquoisePogs: you may want to show the before & after as to what benefit there is to having so many routes, because if you did need all the routes + plugins then you'd need to add /ZLUX/plugins

bin/configure.sh
DivergentEuropeans: oops

lchudinov (approved):

bin/configure.sh
```diff
@@ -27,3 +27,34 @@ export NODE_PATH=../..:../../zlux-server-framework/node_modules:$NODE_PATH
...
+    instanceBaseUrls:
+      - https://${ZOWE_EXPLORER_HOST}:${ZOWE_ZSS_SERVER_PORT}/
+    homePageRelativeUrl:
+    routedServices:
+      - gatewayUrl: api/v1
+        serviceRelativeUrl: 
+    apiInfo:
+      - apiId: org.zowe.zss
+        gatewayUrl: api/v1
+        version: 1.15.0+20200901
```
lchudinov, line 49: Perhaps, there is an environment variable for version?

1000TurquoisePogs (approved):
This is a good first version. We can add more if better variables are found or we have better apiinfo over time.

PR #210, opened 2020-08-25
Agent API: Show limited info when RBAC is disabled

lchudinov:
When RBAC is disabled only these services should be available:

- /server/agent/environment (with limited info)
- /server/agent/services


ifakhrutdinov (approved):
Looks good, but I'd probably move the following to its own function
```c
static bool statusEndPointRequireRBAC(const char *endpoint) {
  return !strcmp(endpoint, "config") ||
         !strcmp(endpoint, "log") ||
         !strcmp(endpoint, "logLevels");
}
```

PR #217, opened 2020-09-24
Remove Node PreReq from ZSS

sakeerthy:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,5 +4,13 @@ All notable changes to the ZSS package will be documented in this file.
 
 ## Recent Changes
 
+## `1.17.0`
+
+### New features and enhancements
+- ZSS no longer requires NodeJS for its configure.sh script
+
+## `1.12.0`
+
+### New features and enhancements
 - Added scripts to allow ZSS to be its own component in the zowe install packaging
 
--- a/bin/configure.sh
+++ b/bin/configure.sh
@@ -55,6 +55,7 @@ if [ ! -f "$currentJsonConfigPath" ]; then
   mkdir -p $SITE_PLUGIN_STORAGE
   mkdir -p $INSTANCE_PLUGIN_STORAGE
   mkdir -m 700 -p $INSTANCE_CONFIG
+  chmod 700 $INSTANCE_CONFIG
   mkdir -p $GROUPS_DIR
   mkdir -p $USERS_DIR
   mkdir -p $PLUGINS_DIR
--- a/bin/configure.sh
+++ b/bin/configure.sh
@@ -56,6 +56,7 @@ if [ ! -f "$currentJsonConfigPath" ]; then
   mkdir -p $INSTANCE_PLUGIN_STORAGE
   mkdir -m 700 -p $INSTANCE_CONFIG
   chmod 700 $INSTANCE_CONFIG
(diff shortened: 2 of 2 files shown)


1000TurquoisePogs (changes requested):
The `sed` commands do not do what is ZSS needs to be done.
What should be done: sed is used to make edits to the default config file to generate a config file that has values specific to the instance
What is actually done: sed edits the correct places by placing the environment variable **KEY** instead of its **VALUE**, and then prints the output to STDOUT instead of a file, such that it prints the entire JSON to the screen about 5 times and each time it only has the last edit.
An example of the problems:
```
  "productDir":"../defaults",
  "siteDir":"../deploy/site",
  "instanceDir":"../deploy/instance",
  "groupsDir":"../deploy/instance/groups",
  "usersDir":"../deploy/instance/users",
  "pluginsDir":"$PLUGINS_DIR",
```
Note it says $PLUGINS_DIR instead of the value of $PLUGINS_DIR

1000TurquoisePogs:
`sed 's@"productDir":"../defaults"@"productDir":"'${PRODUCT_DIR}'"@g' $currentJsonConfigPath`
instead of
`sed 's@"productDir":"../defaults"@"productDir":"${PRODUCT_DIR}"@g' $currentJsonConfigPath`
works better, but still prints to STDOUT.
The difference was I put `'` around the environment var.
However, the environment var value is also wrong, ` PRODUCT_DIR=$PWD../defaults` misses a slash and may be interpreted poorly if something does not process `../` properly. A more robust way is to do 
```
PRODUCT_DIR=$(cd `$PWD/../defaults` && pwd)
```
This technique is used in some of our other scripts.

1000TurquoisePogs:
I'm not a sed master. This is a bit funny looking but works
```
  sed 's@"productDir":"../defaults"@"productDir":"'${PRODUCT_DIR}'"@g' $currentJsonConfigPath > ${currentJsonConfigPath}.zwetmp1
  sed 's@"siteDir":"../deploy/site"@"siteDir":"'${SITE_DIR}'"@g' ${currentJsonConfigPath}.zwetmp1 > ${currentJsonConfigPath}.zwetmp2
  sed 's@"instanceDir":"../deploy/instance"@"instanceDIR":"'${DESTINATION}'"@g' ${currentJsonConfigPath}.zwetmp2 > ${currentJsonConfigPath}.zwetmp1
  sed 's@"groupsDir":"../deploy/instance/groups"@"groupsDir":"'${GROUPS_DIR}'"@g' ${currentJsonConfigPath}.zwetmp1 > ${currentJsonConfigPath}.zwetmp2
  sed 's@"usersDir":"../deploy/instance/users"@"usersDir":"'${USERS_DIR}'"@g' ${currentJsonConfigPath}.zwetmp2 > ${currentJsonConfigPath}.zwetmp1
  sed 's@"pluginsDir":"../defaults/plugins"@"pluginsDir":"'${PLUGINS_DIR}'"@g' ${currentJsonConfigPath}.zwetmp1 > ${currentJsonConfigPath}
  rm ${currentJsonConfigPath}.zwetmp1 ${currentJsonConfigPath}.zwetmp2
```

1000TurquoisePogs (approved):
We have tested that this now has the correct behavior.

PR #222, opened 2020-10-09
Service for mapping public part of the certificate to the MF userid

balhar-jakub:


balhar-jakub:
Copying the information 

We have opened related PRs:
Documentation and setup - zowe/docs-site#1443
Testing - zowe/zowe-install-packaging#1713
Set up for the CICD pipeline: zowe/zowe-install-packaging#1714 and zowe/zowe-install-packaging#1726

As for the API ML the functionality is merged and hidden behind a feature flag.

balhar-jakub:
There is one question that you may also have an answer to and that is whether it is possible to change the port the ZSS runs on within the Zowe. 

We are aware that it is possible to setup via instance.json file for the ZSS running standalone but is there any property within the installation?

1000TurquoisePogs:
>is there any property within the installation
Yes, it is here:
https://github.com/zowe/zowe-install-packaging/blob/master/scripts/instance.template.env#L43

PR #226, opened 2020-10-26
TLS Support

lchudinov:
TLS Support using gskit library. 
Depends on
- https://github.com/zowe/zowe-common-c/pull/199

Related PRs:
- Automate TLS setup for zss zowe/zlux-app-server#151
- TLS support for ZSS agent zowe/zlux-server-framework#271

Linked issue: https://github.com/zowe/zlux/issues/632

### DIFF
--- a/bin/configure.sh
+++ b/bin/configure.sh
@@ -64,8 +64,7 @@ if [ ! -f "$currentJsonConfigPath" ]; then
 fi
 
 APP_WORKSPACE_DIR=${INSTANCE_DIR}/workspace/app-server
-if [ "${ZOWE_ZSS_SERVER_TLS}" = "false" ]
-then
+if [ "${ZOWE_ZSS_SERVER_TLS}" = "false" ]; then
   PROTOCOL="http"
 else
   PROTOCOL="https"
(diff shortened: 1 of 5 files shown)


c/zss.c
```diff
@@ -847,12 +850,82 @@ static void readAgentAddressAndPort(JsonObject *serverConfig, JsonObject *envCon
...
+                                   JsonObject *envConfig,
+                                   char **outAddress,
+                                   int *outPort,
+                                   TlsSettings **outSettings
+                                  ) {
+  int port = jsonObjectGetNumber(envConfig, ENV_AGENT_HTTPS_KEY(PORT_KEY));
+  char *address = jsonObjectGetString(envConfig, ENV_AGENT_HTTPS_KEY(IP_ADDRESSES_KEY));
+
+  TlsSettings *settings = (TlsSettings*)SLHAlloc(slh, sizeof(*settings));
+  settings->keyring = jsonObjectGetString(envConfig, ENV_AGENT_HTTPS_KEY(KEYRING_KEY));
```
1000TurquoisePogs, line 874: does this only support keyrings?
can gskit also read in unix files?

lchudinov: `keyring` is implicit, See for details https://github.com/zowe/zowe-common-c/blob/staging/h/tls.h#L26
I use a `.kdb` gskit database file, for example, see https://github.com/lchudinov/tls-examples

bin/configure.sh
```diff
@@ -99,3 +99,15 @@ iconv -f IBM-1047 -t IBM-850 ${STATIC_DEF_CONFIG_DIR}/zss.ebcidic.yml > $STATIC_
 rm ${STATIC_DEF_CONFIG_DIR}/zss.ebcidic.yml
 chmod 770 $STATIC_DEF_CONFIG_DIR/zss.yml
 fi
+
+# Setup certificates
+PREFIX="ZWED_agent_https_"
```
1000TurquoisePogs, line 104: This new section (and the `instanceBaseUrls` above) needs to be conditional upon var `ZOWE_ZSS_SERVER_TLS=true/false` as seen in instance.env.
This should enable the user to quickly switch between https and http for backward  compatibility.

1000TurquoisePogs: Also, currently the ZSS ip address defaults to 127.0.0.1 and ignores all the default values in instance.env, which confuses users.
When TLS is true, I think it would be great to set ip address to external.

`if (tls && !ZWED_agent_https_ipAddresses) { export ZWED_agent_https_ipAddresses = $ZOWE_IP_ADDRESS }`

1000TurquoisePogs (changes requested):
This is working well without any zowe-install-packaging changes, however I have some suggestions that are important for end users to use effectively.

Also, as expected, changing to HTTPS breaks our tests. You can see the result at https://wash.zowe.org:8443/job/zowe-install-test/job/PR-1996/1/display/redirect

bin/configure.sh
1000TurquoisePogs: https://github.com/zowe/zowe-install-packaging/pull/1996/commits/a11690f3af9b7dc88c6709e75b0e4b10ac94f78c

PR #227, opened 2020-10-30
Enable auth tracing 

achmelo:
Test build to trace JWT auth error


1000TurquoisePogs:
Closing as outdated, since it looks like all this code is in.

PR #240, opened 2020-12-22
APIML Storage

lchudinov:
## APIML Storage

This PR implements Storage API using Cashing Service in APIML.
 
Depends on https://github.com/zowe/zowe-common-c/pull/216

### Testing
This PR adds a ZSS plugin and UI to test Storage API https://github.com/zowe/sample-angular-app/pull/57


1000TurquoisePogs:
Your test case probably works, but I think I see a design issue when handling user requests, here: https://github.com/zowe/zss/pull/242#pullrequestreview-561321442

c/storageApiml.c
```diff
@@ -0,0 +1,648 @@
...
+      receiveResponse(httpClientContext, session, &status);
+    }
+    if (status) {
+      *statusOut = status;
+      break;
+    }
+  } while (0);
+  if (status) {
+    if (session) {
+      httpClientSessionDestroy(session);
```
1000TurquoisePogs, line 312: Looks like this is only being done in the case of failure. Is this correct or a leak?

1000TurquoisePogs: Nevermind, its handled in freeApimlResponse

c/storageApiml.c
```diff
@@ -0,0 +1,648 @@
...
+  }
+  HttpClientSettings *clientSettings = (HttpClientSettings*)safeMalloc(sizeof(*clientSettings), "HTTP Client Settings");
+  if (!clientSettings) {
+    safeFree((char*)storage, sizeof(*storage));
+    safeFree((char*)apimlStorage, sizeof(*apimlStorage));
+    return NULL;
+  }
+  clientSettings->host = settings->host;
+  clientSettings->port = settings->port;
+  clientSettings->recvTimeoutSeconds = 10;
```
1000TurquoisePogs, line 533: 10 seconds should be enough, but shouldnt this be a parameter from somewhere?

c/zss.c
```diff
@@ -988,13 +1032,113 @@ static bool readAgentHttpsSettings(ShortLivedHeap *slh,
...
+#define NODE_MEDIATION_LAYER_SERVER_PREFIX   "ZWED_node_mediationLayer_server_"
+#define ENV_NODE_MEDIATION_LAYER_SERVER_KEY(key) NODE_MEDIATION_LAYER_SERVER_PREFIX key
+
+// Returns false if gateway host and port not found
+static bool readGatewaySettings(JsonObject *serverConfig,
+                                JsonObject *envConfig,
+                                char **outGatewayHost,
+                                int *outGatewayPort
+                               ) {
+  char *gatewayHost = jsonObjectGetString(envConfig, ENV_NODE_MEDIATION_LAYER_SERVER_KEY("hostname"));
```
1000TurquoisePogs, line 1066: gatewayHostname also may exist to separate discovery host from gateway host.

c/zss.c
```diff
@@ -988,13 +1032,113 @@ static bool readAgentHttpsSettings(ShortLivedHeap *slh,
...
+  JsonObject *mediationLayerSettings = jsonObjectGetObject(nodeSettings, "mediationLayer");
+  if (!mediationLayerSettings) {
+    return false;
+  }
+  JsonObject *serverSettings = jsonObjectGetObject(mediationLayerSettings, "server");
+  if (!serverSettings) {
+    return false;
+  }
+  if (!gatewayHost) {
+    gatewayHost = jsonObjectGetString(serverSettings, "hostname");
```
1000TurquoisePogs, line 1087: same here

c/zss.c
lchudinov: Fixed

c/zss.c
lchudinov: Fixed

c/storageApiml.c
lchudinov: I moved the parameter into `APIMLStorageSettings` so that we can make it configurable via some `ZWED_` environment variable or using `server.json` if we need it in the future.

1000TurquoisePogs (approved):

PR #242, opened 2020-12-31
Use APIML and Memory Storages

lchudinov:
Work is progress

Depends on
- https://github.com/zowe/zss/pull/240
- https://github.com/zowe/zss/pull/226
- https://github.com/zowe/zowe-common-c/pull/180
- https://github.com/zowe/zowe-common-c/pull/183


c/zss.c
```diff
@@ -667,8 +676,9 @@ static WebPluginListElt* readWebPluginDefinitions(HttpServer *server, ShortLived
               if (identifier && pluginLocation) {
                 JsonObject *pluginDefinition = readPluginDefinition(slh, identifier, pluginLocation, relativeTo);
                 if (pluginDefinition) {
-                  WebPlugin *plugin = makeWebPlugin(pluginLocation, pluginDefinition, internalAPIMap,
-                                                    &idMultiplier, pluginLogLevel);
+                  Storage *storage = makeMemoryStorage(NULL);
```
1000TurquoisePogs, line 679: This must be conditional.
makeMemoryStorage is for when apiml does not exist. When apiml does exist, need to use makeApimlStorage.

1000TurquoisePogs (commented):
You might have a design issue.

makeApimlStorage needs the apimlAuthenticationToken which is included in every request to ZSS from app-server or other servers/clients. So in order to assign `Storage` at this time it must be ok with being initialized with token=`NULL`. 

Every user has a different apimlAuthenticationToken, so the `Storage` either needs to be able to extract it from every `request`.

I think you need a PR to zowe-common-c to provide the request headers to the `Storage` object right before handling each request, probably around this line: https://github.com/zowe/zowe-common-c/blob/master/c/httpserver.c#L3275

c/zss.c
lchudinov: @1000TurquoisePogs I am uncertain about authentication. What I noticed working with Cashing Service is APIML considers the client of the Cashing Service as a service(not a regular user). Most likely, in case of Caching Service we need Apimltoken which is per service(not per user).

PR #247, opened 2021-01-15
Allow AUX to run as a non-reusable address space

ifakhrutdinov:
#### Overview
Some z/OS services like ones based on non-stacking PC-routines cannot be called from an address space running with `REUSASID=YES`. If the caller attempts doing that, an S0D3 ABEND is received. In order to allow such client code to run inside an AUX address space, the AUX address space needs to be started with `REUSASID=NO`. This pull-request adds a second version of the AUX start guest function which includes a flag argument and one of the flags forces AUX to be started with `REUSASID=NO`.

**WARNING:** running an AUX address space with `REUSASID=NO` **will** leak ASIDs since an AUX address space uses a system LX for its communication PC routine.

### DIFF
--- a/zis-aux/include/aux-manager.h
+++ b/zis-aux/include/aux-manager.h
@@ -72,6 +72,9 @@ typedef struct ZISAUXParm_tag {
   char value[200];
 } ZISAUXParm;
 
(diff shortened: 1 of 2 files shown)


daveyc (approved):
LGTM

Issue #250, opened 2021-01-20
Avoid using AUX communication PC for termination


ifakhrutdinov:
#### Problem description
The AUX address space can run with `REUSASID=NO`. When it does, it will leak the ASID every time it's terminated because it uses a system LX for its communication PC. Using a non-system LX is not an option because once the parent (ZIS) issues `ETCON`, the AUX address space will be converted to non-reusable, again, leading to leaking the ASID.

#### Proposed solution
Since the communication PC isn't used much, we should introduce an option to disable it to avoid leaking ASIDs. Disabling the PC will disrupt modify command, work and termination signals sent from ZIS to AUX. The termination signal is all we care at the moment so we need to add an additional communication path for the termination signal. This new path is an ECB on which AUX will wait in a dedicated thread. Once ZIS is terminated, it will post on that ECB instead of calling the now disabled PC routine.

PR #251, opened 2021-01-21
AUX communication PC disablement

ifakhrutdinov:
#### Overview

With the introduction of the REUSASID=NO option for AUX, it became somewhat dangerous to run it with the communication PC as it would leak ASIDs. This commit adds the ability to disable the communication PC and introduces an alternative communication path for the termination signal.

##### Major changes:
* Disablement of the communication PC
* Termination signal via a cross-memory POST

##### Minor changes:
* Defines for AUX parameters
* Communication area versioning

Fixes: #250

### DIFF
--- a/zis-aux/include/aux-manager.h
+++ b/zis-aux/include/aux-manager.h
@@ -47,7 +47,10 @@ typedef struct ZISAUXManager_tag {
 
 typedef struct ZISAUXCommArea_tag {
   char eyecatcher[8];
-#define ZISAUX_COMM_EYECATCHER    "ZISAUXC "
+#define ZISAUX_COMM_EYECATCHER    "ZISAUXCA"
+  uint8_t version;
+#define ZISAUX_COMM_VERSION 2
+  char reserved0[3];
   int32_t flag;
 #define ZISAUX_HOST_FLAG_READY          0x00000001
 #define ZISAUX_HOST_FLAG_TERMINATED     0x00000002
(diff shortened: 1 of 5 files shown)


zis-aux/include/aux-manager.h
```diff
@@ -51,13 +51,18 @@ typedef struct ZISAUXCommArea_tag {
...
 #define ZISAUX_HOST_FLAG_TERMINATED     0x00000002
+#define ZISAUX_HOST_FLAG_COMM_PC_ON     0x00000004
   uint32_t pcNumber;
   uint32_t sequenceNumber;
   int32_t termECB;
   uint64_t stoken;
   ASCB * __ptr32 ascb;
   uint16_t parentASID;
-  char reserved0[2];
+  uint8_t version;
```
daveyc, line 61: It's unfortunate the the version is positioned at the bottom of the structure. Can this be rolled into the EYECATHER?

ifakhrutdinov: Yeah, we can take a risk and make this incompatible with the previous version... Let me think about this.

ifakhrutdinov: Okay, I re-worked the communication area and changed the eyecatcher. This way the worst thing which may happen, if the old version is used, is the eyecatcher check will fail and the AUX A/S will terminate.

ifakhrutdinov:
@daveyc I've made another change - I had to move the code which starts the communication TCB. The TCB would use the comm area from the context, but that field was initialized **after** the TCB has been started. So there was a race condition and in come cases the wait call would ABEND because the ECB was at address 0x30.

daveyc (approved):
Much better

daveyc (approved):

PR #254, opened 2021-01-29
Fix a null pointer dereference

ifakhrutdinov:
#### Overview
In case a bad parm is specified in ZIS config, the code may end up dereferencing a NULL pointer. This fix makes the default value an empty string.

### DIFF
--- a/c/zis/parm.c
+++ b/c/zis/parm.c
@@ -439,7 +439,7 @@ int zisReadMainParms(ZISParmSet *parms, const ZISMainFunctionParms *mainParms) {
     }
 
     char *key = workBuffer;
-    char *value = NULL;
+    char *value = "";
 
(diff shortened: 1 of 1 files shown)


daveyc (commented):
Should `zisPutParmValue` also be changed to handle null pointers?

ifakhrutdinov:
> 
> 
> Should `zisPutParmValue` also be changed to handle null pointers?

I think it should be documented that it doesn't accept `NULL`. Or it needs to handle `NULL` for all the parameters, not sure where to draw the line. I'll just document it probably for now.

daveyc:
IMO `NULL` checking should be done at every stage of the stack but I will approve this.

daveyc (approved):

PR #269, opened 2021-03-18
Make sure the AUX API is backward and forward compatible with 1.19.

ifakhrutdinov:
Clone of https://github.com/zowe/zss/pull/268


1000TurquoisePogs (approved):
Testing has shown this to be a good fix for backward compatibility

PR #278, opened 2021-04-10
add swagger url to static registeration

NakulManchanda:
Signed-off-by: Nakul Manchanda <nakul.manchanda@ibm.com>

### DIFF
--- a/apiml-static-reg.yaml.template
+++ b/apiml-static-reg.yaml.template
@@ -12,7 +12,7 @@ services:
     apiInfo:
       - apiId: org.zowe.zss
         gatewayUrl: api/v1
-        swaggerUrl: https://${ZOWE_EXPLORER_HOST}:${ZOWE_ZLUX_SERVER_HTTPS_PORT:-8544}/api-docs/zss
+        swaggerUrl: https://${ZOWE_EXPLORER_HOST}:${ZOWE_ZLUX_SERVER_HTTPS_PORT:-8544}/api-docs/agent
         # documentationUrl: TODO
 catalogUiTiles:
   zss:
--- a/apiml-static-reg.yaml.template
+++ b/apiml-static-reg.yaml.template
@@ -12,7 +12,7 @@ services:
     apiInfo:
       - apiId: org.zowe.zss
         gatewayUrl: api/v1
-        # swaggerUrl: TODO
+        swaggerUrl: https://${ZOWE_EXPLORER_HOST}:${ZOWE_ZLUX_SERVER_HTTPS_PORT}/api-docs/zss
         # documentationUrl: TODO
 catalogUiTiles:
   zss:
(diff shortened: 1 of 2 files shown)


apiml-static-reg.yaml.template
```diff
@@ -12,7 +12,7 @@ services:
     apiInfo:
       - apiId: org.zowe.zss
         gatewayUrl: api/v1
-        # swaggerUrl: TODO
+        swaggerUrl: https://${ZOWE_EXPLORER_HOST}:${ZOWE_ZLUX_SERVER_HTTPS_PORT}/api-docs/zss
```
1000TurquoisePogs, line 15: this would probably fail if zlux was not running, which is possible. could we use some env var to make it undefined in that case?

NakulManchanda: Sure, we can make the whole url undefined if thats the case. Np

NakulManchanda: @1000TurquoisePogs if swagger is unreachable it wont show up, we should not corrupt the file though i believe

NakulManchanda: made 8544 as default

timgerstel (approved):

DivergentEuropeans (changes requested):
For https://github.com/zowe/zlux-server-framework/pull/279 & #278, if
I'm seeing all endpoints successfully so things lgtm. :white_check_mark:
However, the swaggers docs for things like
https://mainframe:8544/api-docs/agent/server-plugins-api
is the same one for App server, which isn't entirely accurate for ZSS. For example, it tells me I can do,
{zssHost}:{zssPort}/server to get list of server endpoints but I cannot. I can however do
{zssHost}:{zssPort}/server/agent
So it seems like your code is good and it makes sense, but it's going to add Swagger docs to ZSS that aren't 100% accurate to ZSS. So before this to be merged (or at least before 278 should be merged which says ZSS Swagger is https://mainframe:8544/api-docs/agent) then we'd want to make separate Swagger docs for ZSS. This shouldn't be a long process maybe, I don't think they differ by much

apiml-static-reg.yaml.template
```diff
@@ -12,7 +12,7 @@ services:
     apiInfo:
       - apiId: org.zowe.zss
         gatewayUrl: api/v1
-        # swaggerUrl: TODO
+        swaggerUrl: https://${ZOWE_EXPLORER_HOST}:${ZOWE_ZLUX_SERVER_HTTPS_PORT:-8544}/api-docs/agent
```
1000TurquoisePogs, line 15: /server/agent/swagger as per comments in https://github.com/zowe/zlux-server-framework/pull/279

1000TurquoisePogs (changes requested):
I think the URL scheme should be changed to align with pre-existing URLs.
We have /server/agent
Why not /server/agent/swagger and /server/agent/swagger/unixfile , /server/agent/swagger/datasetContents, etc...

apiml-static-reg.yaml.template
NakulManchanda: `/server/agent` endpoints seems protected via rbac setting
image

1000TurquoisePogs (approved):

PR #279, opened 2021-04-12
Bugfix for missing xmem property reading

1000TurquoisePogs:
app-server does this already and their environments used to be shared but we're moving away from that model so this is a bug waiting to happen. Fixing it before it becomes an issue.


1000TurquoisePogs:
See the original code: https://github.com/zowe/zlux-app-server/blob/staging/bin/convert-env.sh#L200

PR #281, opened 2021-04-13
Initial commit for SAF query interpretation methods

DivergentEuropeans:
Similar to what zlux-server-framework\plugins\sso-auth\lib\safprofile.js does

Turns a SAF URL into a SAF query i.e. /plugins GET undefined
->
ZLUX.0.COR.GET.PLUGINS

ZSS needs to be changed to use this internally

Signed-off-by: Leanid Astrakou <lastrakou@rocketsoftware.com>

### DIFF
--- a/c/authService.c
+++ b/c/authService.c
@@ -19,6 +19,7 @@
 #include <iconv.h>
 #include <dirent.h>
 #include <pthread.h>
+#include <regex.h> 
 
 #include "authService.h"
 #include "zowetypes.h"
@@ -66,6 +67,19 @@
 
 static int serveAuthCheck(HttpService *service, HttpResponse *response);
 
+const char* getProfileNameFromRequest(char *url, const char *method, int instanceID);
+
+const char* makeProfileName(
+  char *type,
+  char *productCode, 
+  int instanceID, 
+  char *pluginID, 
+  char *rootServiceName, 
+  char *serviceName,
+  char *method,
+  char *scope,
+  char subUrl[12][50]);
+
 int installAuthCheckService(HttpServer *server) {
 //  zowelog(NULL, 0, ZOWE_LOG_DEBUG2, "begin %s\n",
 //  __FUNCTION__);
@@ -74,6 +88,7 @@ int installAuthCheckService(HttpServer *server) {
   httpService->authType = SERVICE_AUTH_NATIVE_WITH_SESSION_TOKEN;
   httpService->serviceFunction = &serveAuthCheck;
   httpService->runInSubtask = FALSE;
+  getProfileNameFromRequest("/plugins", "GET", -1);
   registerHttpService(server, httpService);
 //  zowelog(NULL, 0, ZOWE_LOG_DEBUG2, "end %s\n",
 //  __FUNCTION__);
@@ -173,15 +188,207 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
     respondWithError(res, HTTP_STATUS_BAD_REQUEST, "Unexpected access level");
     return 0;
   }
-  /* printf("query: user %s, class %s, entity %s, access %d\n", userName, class,
-      entity, access); */
+  // printf("\n\nquery: user %s, class %s, entity %s, access %d accessStr %s\n", userName, class,
+      // entity, access, accessStr);
   privilegedServerName = getConfiguredProperty(service->server,
       HTTP_SERVER_PRIVILEGED_SERVER_PROPERTY);
   rc = zisCheckEntity(privilegedServerName, userName, class, entity, access,
       &reqStatus);
+      printf("\n\nprivileged server name: %s", privilegedServerName);
   respond(res, rc, &reqStatus);
   return 0;
 }
+
+const char* getProfileNameFromRequest(char *url, char *method, int instanceID) {
+  char type[8] = "NULL"; // core || config || service
+  char productCode[50] = "NULL";
+  char rootServiceName[50] = "NULL";
+  char subUrl[12][50];
+  char profileName[150];
+  char scope[50];
+  char _p[50] = "NULL", pluginID[50] = "NULL", _s[50] = "NULL", serviceName[50] = "NULL", _v[50] = "NULL";
+  char regexStr[] = "^/[A-Za-z0-9]*/plugins/";
+  
+  regex_t regex;
+  int value;
+  value = regcomp(&regex, regexStr, REG_EXTENDED);
+
+  if (value != 0) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
+           "RegEx compiled successfully.");
+  } else {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
+           "RegEx compilation error %s.", regexStr);
+  }
+  value = regexec(&regex, url, 0, NULL, 0);
+  char urlCpy[150];
+  strcpy(urlCpy, url);
+  int index = 0;
+  while (urlCpy[index]) { // Capitalize query
+      urlCpy[index] = toupper(urlCpy[index]);
+      index++;
+  }
+  if (instanceID < 0) { // Set instanceID
+    instanceID = 0;
+  }
+  if (value == REG_NOMATCH) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
+           "RegEx didn't match.");
+    char * token = strtok(urlCpy, "/");
+    int subUrlIndex = -1;
+    while( token != NULL ) {
+      if (strcmp(rootServiceName, "NULL") == 0)
+      {
+        strcpy(rootServiceName, token);
+      } else {
+        strcpy(subUrl[subUrlIndex], token);
+      }
+      subUrlIndex++;
+      token = strtok(NULL, "/");
+    }
+    strcpy(productCode, "ZLUX");
+    strcpy(type, "core");
+  }
+  else if (!value) {
+    char * token = strtok(urlCpy, "/");
+    int subUrlIndex;
+    subUrlIndex = 0;
+    while( token != NULL ) {
+      switch(subUrlIndex) {
+        case 0:
+          strcpy(productCode, token);
+          break;
+        case 1:
+          strcpy(_p, token);
+          break;
+        case 2:
+          strcpy(pluginID, token);
+          break;
+        case 3:
+          strcpy(_s, token);
+          break;
+        case 4:
+          strcpy(serviceName, token);
+          break;
+        case 5:
+          strcpy(_v, token);
+          break;
+        default:
+          strcpy(subUrl[subUrlIndex-6], token); // subtract 6 from maximum index to begin init subUrl array at 0
+      }
+      
+      subUrlIndex++;
+      token = strtok(NULL, "/");
+    }
+    if ((strcmp(pluginID, "ORG.ZOWE.CONFIGJS") == 0) && (strcmp(serviceName, "DATA") == 0))
+    {
+      strcpy(type, "config");
+      strcpy(pluginID, subUrl[0]);
+      strcpy(scope, subUrl[1]);
+      
+    } else {
+      strcpy(type, "service");
+    }
+    char* ch; 
+    char* chReplace;
+    ch = ".";
+    chReplace = "_";
+    for (index = 0; index <= strlen(pluginID); index++)
+  	{
+  		if (pluginID[index] == *ch)  
+      {
+        pluginID[index] = *chReplace;
+      }
+    }
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
+           "RegEx match OK.");
+  }
+  else {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+           "RegEx match failed.");
+  }
+  strcpy(profileName, makeProfileName(type, 
+    productCode, 
+    instanceID, 
+    pluginID, 
+    rootServiceName,
+    serviceName,
+    method,
+    scope,
+    subUrl));
+  // printf("\n\nFinal query profileName & URL %s - %s\n\n", profileName, url);
+  
+  /* Free memory allocated to the pattern buffer by regcomp() */
+  regfree(&regex);
+  return profileName;
+}
+
+const char* makeProfileName(
+  char *type,
+  char *productCode, 
+  int instanceID, 
+  char *pluginID, 
+  char *rootServiceName, 
+  char *serviceName,
+  char *method,
+  char *scope,
+  char subUrl[15][50]) {
+  char profileName[200] = "";
+  if (strcmp(productCode, "NULL") == 0) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+           "Broken SAF query. Missing product code.");
+    return "NULL";
+  }
+  if (instanceID == -1) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+           "Broken SAF query. Missing instance ID.");
+    return "NULL";
+  }
+  if (strcmp(method, "NULL") == 0) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+           "Broken SAF query. Missing method.");
+    return "NULL";
+  }
+  // char someString[50] = { strcpy(*someString, type) };
+  if (strcmp(type, "service") == 0) {
+    if (strcmp(pluginID, "NULL") == 0) {
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing plugin ID.");
+      return "NULL";
+    }
+    if (strcmp(serviceName, "NULL") == 0) {
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing service name.");
+      return "NULL";
+    }
+    snprintf(profileName, 1024, "%s.%d.SVC.%s.%s.%s", productCode, instanceID, pluginID, serviceName, method);
+    return profileName;
+    
+  } else if (strcmp(type, "config") == 0) {
+    if (strcmp(pluginID, "NULL") == 0) {
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing plugin ID.");
+      return "NULL";
+    }
+    if (strcmp(scope, "NULL") == 0) {
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing scope.");
+      return "NULL";
+    }
+    snprintf(profileName, 1024, "%s.%d.CFG.%s.%s.%s", productCode, instanceID, pluginID, method, scope); 
+    return profileName;
+  } else if (strcmp(type, "core") == 0) {
+    if (strcmp(rootServiceName, "NULL") == 0) {
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing root service name.");
+      return "NULL";
+    }
+    snprintf(profileName, 1024, "%s.%d.COR.%s.%s", productCode, instanceID, method, rootServiceName); 
+    return profileName;
+  }
+}
+
+/* Method goes here to do the same thing serveAuthCheck is doing except w/o input HttpService */
 
 void respondWithJsonStatus(HttpResponse *response, const char *status, int statusCode, const char *statusMessage) {
     jsonPrinter *out = respondWithJsonPrinter(response);


c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing plugin ID.");
+      return "NULL";
+    }
+    if (strcmp(scope, "NULL") == 0) {
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing scope.");
+      return "NULL";
+    }
+    snprintf(profileName, 1024, "%s.%d.CFG.%s.%s.%s", productCode, instanceID, pluginID, method, scope);
```
lchudinov, line 378: `profileName` can hold 250 chars at max, why does 1024 go here? It's a possible buffer overflow.

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
       HTTP_SERVER_PRIVILEGED_SERVER_PROPERTY);
   rc = zisCheckEntity(privilegedServerName, userName, class, entity, access,
       &reqStatus);
+      printf("\n\nprivileged server name: %s", privilegedServerName);
   respond(res, rc, &reqStatus);
   return 0;
 }
 
+const char* getProfileNameFromRequest(char *url, char *method, int instanceID) {
+  char type[8] = "NULL"; // core || config || service
```
lchudinov, line 203: why to use `"NULL"` string everywhere, wouldn't an empty string work?

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  char productCode[50] = "NULL";
+  char rootServiceName[50] = "NULL";
+  char subUrl[15][50];
+  char profileName[250];
+  char scope[50];
+  char _p[50] = "NULL", pluginID[50] = "NULL", _s[50] = "NULL", serviceName[50] = "NULL", _v[50] = "NULL";
```
lchudinov, line 209: `"NULL"`

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  }
+  if (value == REG_NOMATCH) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
+           "RegEx didn't match.");
+    char * token = strtok(urlCpy, "/");
+    int subUrlIndex = -1;
+    while( token != NULL ) {
+      if (strcmp(rootServiceName, "NULL") == 0)
+      {
+        strcpy(rootServiceName, token);
```
lchudinov, line 242: Buffer overflow is possible.

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+      } else {
+        strcpy(subUrl[subUrlIndex], token);
```
lchudinov, line 244: Buffer overflow is possible.

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+    strcpy(type, "core");
+  }
+  else if (!value) {
+    char * token = strtok(urlCpy, "/");
+    int subUrlIndex;
+    subUrlIndex = 0;
+    while( token != NULL ) {
+      switch(subUrlIndex) {
+        case 0:
+          strcpy(productCode, token);
```
lchudinov, line 259: Buffer overflow is possible.

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+          break;
+        case 1:
+          strcpy(_p, token);
+          break;
+        case 2:
+          strcpy(pluginID, token);
```
lchudinov, line 265: Buffer overflow is possible.

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+          break;
+        case 3:
+          strcpy(_s, token);
+          break;
+        case 4:
+          strcpy(serviceName, token);
```
lchudinov, line 271: Buffer overflow is possible.

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+          break;
+        case 5:
+          strcpy(_v, token);
```
lchudinov, line 274: Buffer overflow is possible.

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+          break;
+        default:
+          strcpy(subUrl[subUrlIndex-6], token); // subtract 6 from maximum index to begin init subUrl array at 0
```
lchudinov, line 277: Buffer overflow is possible.

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+    rootServiceName,
+    serviceName,
+    method,
+    scope,
+    subUrl));
+  // printf("\n\nFinal query profileName & URL %s - %s\n\n", profileName, url);
+  
+  /* Free memory allocated to the pattern buffer by regcomp() */
+  regfree(&regex);
+  return profileName;
```
lchudinov, line 323: The variable profileName is allocated on the stack so its lifetime is limited to the enclosing function. If you want to return profileName you have to allocate it with safeMalloc and SLHAlloc

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  char profileName[250] = "";
+  if (strcmp(productCode, "NULL") == 0) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+           "Broken SAF query. Missing product code.");
+    return "NULL";
+  }
+  if (instanceID == -1) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+           "Broken SAF query. Missing instance ID.");
+    return "NULL";
```
lchudinov, line 345: wouldn't NULL pointer work here?

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+           "Broken SAF query. Missing method.");
+    return "NULL";
+  }
+  // char someString[50] = { strcpy(*someString, type) };
+  if (strcmp(type, "service") == 0) {
+    if (strcmp(pluginID, "NULL") == 0) {
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing plugin ID.");
+      return "NULL";
```
lchudinov, line 357: ?

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+    }
+    if (strcmp(serviceName, "NULL") == 0) {
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing service name.");
+      return "NULL";
```
lchudinov, line 362: ?

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+    }
+    snprintf(profileName, 1024, "%s.%d.SVC.%s.%s.%s", productCode, instanceID, pluginID, serviceName, method);
```
lchudinov, line 364: 1024 > 250. buffer overflow is possible

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+    return profileName;
+    
+  } else if (strcmp(type, "config") == 0) {
+    if (strcmp(pluginID, "NULL") == 0) {
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing plugin ID.");
+      return "NULL";
```
lchudinov, line 371: ?

c/authService.c
lchudinov, line 378: A possible buffer overflow.

c/authService.c
```diff
@@ -173,16 +188,208 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+       "Broken SAF query. Missing scope.");
+      return "NULL";
+    }
+    snprintf(profileName, 1024, "%s.%d.CFG.%s.%s.%s", productCode, instanceID, pluginID, method, scope); 
+    return profileName;
+  } else if (strcmp(type, "core") == 0) {
+    if (strcmp(rootServiceName, "NULL") == 0) {
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing root service name.");
+      return "NULL";
```
lchudinov, line 384: "NULL"

lchudinov (changes requested):

c/authService.c
DivergentEuropeans: I wanted to distinguish an intentionally empty parameter "" vs N/A "NULL"

c/authService.c
DivergentEuropeans: I used "NULL" vs "" because I wanted to differentiate between an empty parameter vs no parameter. I'm not sure of a usecase in the field, but I wanted to have a way for a user URL to distinguish between a parameter that's intentionally empty vs N/A

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
   }
-  /* printf("query: user %s, class %s, entity %s, access %d\n", userName, class,
-      entity, access); */
+  // printf("\n\nquery: user %s, class %s, entity %s, access %d accessStr %s\n", userName, class,
+      // entity, access, accessStr);
   privilegedServerName = getConfiguredProperty(service->server,
       HTTP_SERVER_PRIVILEGED_SERVER_PROPERTY);
   rc = zisCheckEntity(privilegedServerName, userName, class, entity, access,
       &reqStatus);
+      printf("\n\nprivileged server name: %s", privilegedServerName);
```
timgerstel, line 197: Dont forget to remove printf

timgerstel (commented):
Seems okay to me

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+      }
+    }
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
+           "RegEx match OK.");
+  }
+  else {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+           "RegEx match failed.");
+  }
+  snprintf(profileName, sizeof(profileName), makeProfileName(type,
```
lchudinov, line 310: `sizeof(profileName)` is 4

c/authService.c
```diff
@@ -74,6 +88,7 @@ int installAuthCheckService(HttpServer *server) {
   httpService->authType = SERVICE_AUTH_NATIVE_WITH_SESSION_TOKEN;
   httpService->serviceFunction = &serveAuthCheck;
   httpService->runInSubtask = FALSE;
+  // getProfileNameFromRequest("/plugins", "GET", -1);
```
ifakhrutdinov, line 91: Please don't commit more dead code.

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
     respondWithError(res, HTTP_STATUS_BAD_REQUEST, "Unexpected access level");
     return 0;
   }
-  /* printf("query: user %s, class %s, entity %s, access %d\n", userName, class,
-      entity, access); */
+  // printf("\n\nquery: user %s, class %s, entity %s, access %d accessStr %s\n", userName, class,
```
ifakhrutdinov, line 191: Dead code.

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+      // entity, access, accessStr);
   privilegedServerName = getConfiguredProperty(service->server,
       HTTP_SERVER_PRIVILEGED_SERVER_PROPERTY);
   rc = zisCheckEntity(privilegedServerName, userName, class, entity, access,
       &reqStatus);
+      // printf("\n\nprivileged server name: %s", privilegedServerName);
```
ifakhrutdinov, line 197: Dead code.

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
   respond(res, rc, &reqStatus);
   return 0;
 }
 
+const char* getProfileNameFromRequest(char *url, char *method, int instanceID) {
+  char type[8]; // core || config || service
+  char productCode[1024];
+  char rootServiceName[1024];
+  char subUrl[15][1024];
+  char *profileName = (char*) safeMalloc(1024, "profileName");
```
ifakhrutdinov, line 207: * No need to case
* A null check is needed

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  char scope[1024];
+  char _p[1024], pluginID[1024], _s[1024], serviceName[1024], _v[1024];
```
ifakhrutdinov, line 209: Can you use more descriptive names? What's `_p`?

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  
+  regex_t regex;
+  int value;
+  value = regcomp(&regex, regexStr, REG_EXTENDED);
+
+  if (value != 0) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
+           "RegEx compiled successfully.");
+  } else {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
```
ifakhrutdinov, line 220: Why do you continue if there is an error?

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+           "RegEx compilation error %s.", regexStr);
+  }
+  value = regexec(&regex, url, 0, NULL, 0);
+  char urlCpy[1024];
```
ifakhrutdinov, line 224: * Can an URL be larger that the buffer? If it is, will it be just truncated?
* You have a lot of magic numbers in here. Can you tie them to some `const` variable or a `#define`?

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  snprintf(urlCpy, 1024, url);
+  int index = 0;
+  while (urlCpy[index]) { // Capitalize query
```
ifakhrutdinov, line 227: Helper function?

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  if (instanceID < 0) { // Set instanceID
+    instanceID = 0;
+  }
+  if (value == REG_NOMATCH) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
+           "RegEx didn't match.");
+    char * token = strtok(urlCpy, "/");
+    int subUrlIndex = -1;
+    while( token != NULL ) {
+      if (strcmp(rootServiceName, NULL) == 0)
```
ifakhrutdinov, line 240: Please don't pass `NULL` here.

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+      {
+        snprintf(rootServiceName, 1024, token);
+      } else {
+        snprintf(subUrl[subUrlIndex], 1024, token);
```
ifakhrutdinov, line 244: What if you have more than 15 tokens? Buffer overflow?

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+      urlCpy[index] = toupper(urlCpy[index]);
+      index++;
+  }
+  if (instanceID < 0) { // Set instanceID
+    instanceID = 0;
+  }
+  if (value == REG_NOMATCH) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
+           "RegEx didn't match.");
+    char * token = strtok(urlCpy, "/");
```
ifakhrutdinov, line 237: `strtok()` is not thread-safe.

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  char *type,
+  char *productCode, 
+  int instanceID, 
+  char *pluginID, 
+  char *rootServiceName, 
+  char *serviceName,
+  char *method,
+  char *scope,
+  char subUrl[15][1024]) {
+  char *profileName = (char*) safeMalloc(1024, "profileNameInner");
```
ifakhrutdinov, line 337: NULL check

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  if (strcmp(productCode, NULL) == 0) {
```
ifakhrutdinov, line 338: Lots of undefined behavior with `strcmp` and `NULL`.

c/authService.c
```diff
@@ -66,6 +67,19 @@
 
 static int serveAuthCheck(HttpService *service, HttpResponse *response);
 
+const char* getProfileNameFromRequest(char *url, const char *method, int instanceID);
+
+const char* makeProfileName(
+  char *type,
```
ifakhrutdinov, line 73: If you don't change the memory behind these pointer, make them `const` to better show the intentions.

c/authService.c
```diff
@@ -173,16 +188,209 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+    subUrl));
+  // printf("\n\nFinal query profileName & URL %s - %s\n\n", profileName, url);
+  
+  /* Free memory allocated to the pattern buffer by regcomp() */
+  regfree(&regex);
+  
+  return profileName;
+}
+
+const char* makeProfileName(
```
ifakhrutdinov, line 327: Is this function going to be used outside this source file? If not, make it static.

ifakhrutdinov (changes requested):
@DivergentEuropeans, please see my comments.

In addition to the comments:
* Please add some documentation (preferably Doxygen) to the function prototypes, it helps to understand what they try to achieve. For example, https://github.com/zowe/zowe-common-c/blob/master/h/radmin.h#L520 ✔️ 
* Consider splitting `getProfileNameFromRequest`, it's huge and hard to follow ✔️ 
* You don't need to reinvent all that URL parsing, the HTTP request already has a split URL, see https://github.com/zowe/zss/blob/staging/c/securityService.c#L579. Can you use or re-use any of that? ✔️

c/authService.c
DivergentEuropeans: This was there in staging. Want me to keep?

c/authService.c
DivergentEuropeans: Wdym by null check? I don't see this in the other usages of safeMalloc, sorry new to C

c/authService.c
DivergentEuropeans: URL should practically not be longer than 1024 char's but if it is, it should get truncated safely by methods like snprintf

c/authService.c
DivergentEuropeans: Sorry, saw strtok() in envService.c

is strchr() okay?

DivergentEuropeans:
Needs CHANGELOG

h/authService.h
```diff
@@ -28,6 +28,10 @@
 
 int installAuthCheckService(HttpServer *server);
 void installZosPasswordService(HttpServer *server);
+const char* getProfileNameFromRequest(char *profileName, char *url, char *method, int instanceID);
+int serveAuthCheck(HttpService *service, HttpResponse *res);
```
DivergentEuropeans, line 32: TODO: serveAuthCheck doesn't need to be made public, but is here now (I was testing it) due to Leonty's request to see code earlier

DivergentEuropeans:
TODO: This does not yet have the boolean check for "rbac" property in server config, so all the logic needs to be made conditional

DivergentEuropeans:
Still need to address Irek's code quality comments but ready for review

c/authService.c
DivergentEuropeans: strtok_r I used

c/authService.c
DivergentEuropeans: Made index check

c/authService.c
DivergentEuropeans: well it's 4 lines

DivergentEuropeans: But I'll make other helper functions for some other stuff to shorten getProfileNameFromRequest

c/zss.c
```diff
@@ -317,10 +319,38 @@ static void setPrivilegedServerName(HttpServer *server, JsonObject *mvdSettings,
...
+  int rc = 0;
+  char profileName[ZOWE_PROFILE_NAME_LEN+1] = {0};
+  char method[16];
+  snprintf(method, sizeof(method), "%s", request->method);
+  destructivelyNativize(method);
+  rc = getProfileNameFromRequest(profileName, request->parsedFile, method, -1, response);
+  if (rc != 0) {
+    return FALSE;
+  }
+  rc = serveAuthCheckByParams(service, request->username, "ZOWE", profileName, 2, ENV_SETTINGS);
```
ifakhrutdinov, line 332: What does `2` (and `-1` in the previous call) mean? Why is it not an enum or `#define`?

c/zss.c
```diff
@@ -317,10 +319,38 @@ static void setPrivilegedServerName(HttpServer *server, JsonObject *mvdSettings,
...
+  return TRUE;
+}
+
+/* Future custom ZSS authorization handlers go here */
+static void initializeAuthHandlers(HttpServer *server) {
+  
+  /* NATIVE_WITH_SESSION_TOKEN */ 
+  server->authHandler[0] = (HttpAuthHandler*)safeMalloc31(sizeof(HttpAuthHandler),"HttpAuthHandler");
+  server->authHandler[0]->type = SERVICE_AUTH_NATIVE_WITH_SESSION_TOKEN;
+  server->authHandler[0]->authFunction = &nativeWithSessionTokenAuth;
```
ifakhrutdinov, line 345: It's better not to change the internals of the HTTP server by hand, you can expose this functionality via a public function in `zowe-common-c/h/httpserver.h`.

Maybe something like this?
```c
int httpServerAddAuthHandler(HttpServer *self, int type, AuthHandle *handler);
```

c/authService.c
```diff
@@ -66,6 +68,27 @@
 
 static int serveAuthCheck(HttpService *service, HttpResponse *response);
 
+int serveAuthCheckByParams(HttpService *service, char *userName, char *class, char* entity, int access, JsonObject *envSettings);
```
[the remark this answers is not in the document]
DivergentEuropeans: Addressed in Leonty's code

h/authService.h
```diff
@@ -26,9 +26,37 @@
...
+/**
+ * @brief The function uses makeProfileName function to generate profile name for SAF query
+ * @param profileName Generated profile name goes here
+ * @param parsedFile Refers to the StringList object which contains URL stripped of args
+ * @param instanceID Refers to instanceID for query. If none specified, or negative, then 0
+ * @param HttpResponse Describes the HttpResponse object to return if error encountered
+ *
+ * @return Return non-zero if error
+ */
+int getProfileNameFromRequest(char *profileName, StringList *parsedFile, char *method, int instanceID, HttpResponse *response);
```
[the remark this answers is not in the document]
DivergentEuropeans: Addressed in Leonty's code

c/authService.c
```diff
@@ -173,16 +196,223 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  char scope[STRING_BUFFER_SIZE];
+  char pluginID[STRING_BUFFER_SIZE]; 
+  char serviceName[STRING_BUFFER_SIZE];
+  char urlSegment[STRING_BUFFER_SIZE];
+  int subUrlIndex = 0;
+  
+  snprintf(urlSegment, STRING_BUFFER_SIZE, "%s", stringListPrint(parsedFile, 1, 1, "/", 0));
+  StringListElt *pathSegment = firstStringListElt(parsedFile);
+
+  strupcase(urlSegment);
```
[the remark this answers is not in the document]
DivergentEuropeans: Yes

c/authService.c
```diff
@@ -173,16 +196,223 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  char productCode[STRING_BUFFER_SIZE];
+  char rootServiceName[STRING_BUFFER_SIZE];
+  char subUrl[SAF_SUB_URL_SIZE][STRING_BUFFER_SIZE] = {0};
+  char scope[STRING_BUFFER_SIZE];
+  char pluginID[STRING_BUFFER_SIZE]; 
+  char serviceName[STRING_BUFFER_SIZE];
+  char urlSegment[STRING_BUFFER_SIZE];
+  int subUrlIndex = 0;
+  
+  snprintf(urlSegment, STRING_BUFFER_SIZE, "%s", stringListPrint(parsedFile, 1, 1, "/", 0));
```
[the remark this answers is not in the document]
DivergentEuropeans: Makes sense, good point

c/authService.c
```diff
@@ -173,16 +196,223 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+           "parsedFile urlSegment check didn't match.");
+    subUrlIndex = -1;
+    while (pathSegment != NULL) {
+      snprintf(urlSegment, STRING_BUFFER_SIZE, "%s", pathSegment->string);
+      strupcase(urlSegment);
+      if (rootServiceName == NULL)
+      {
+        snprintf(rootServiceName, STRING_BUFFER_SIZE, urlSegment);
+      } else {
+        if (subUrlIndex < SAF_SUB_URL_SIZE) {
```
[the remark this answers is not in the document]
DivergentEuropeans: Here is a link to a Zowe doc regarding RBAC query best practices: https://docs.zowe.org/stable/extend/extend-desktop/mvd-dataservices/#using-dataservices-with-rbac

Currently there's no hard limit on URL subsections, but there is a hard char limit (after when reached, we just trim instead of rejecting) so in all likelihood it's unlikely for there to be more than 16 (SAF_SUB_URL_SIZE) subsections but I can bump it up to 32 to avoid any problems. When it's larger than SAF_SUB_URL_SIZE, we don't need to handle this case because we just stop adding URL subsections to profile name

c/authService.c
```diff
@@ -173,16 +196,223 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+  strupcase(urlSegment);
+  if (instanceID < 0) { // Set instanceID
+    instanceID = 0;
+  }
+  if (strcmp(urlSegment, "PLUGINS") != 0) {
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
+           "parsedFile urlSegment check didn't match.");
+    subUrlIndex = -1;
+    while (pathSegment != NULL) {
+      snprintf(urlSegment, STRING_BUFFER_SIZE, "%s", pathSegment->string);
```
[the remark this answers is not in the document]
DivergentEuropeans: Yes, according to the Zowe docs, that is the expected behavior therefore an attempt should be made to keep URL's relatively short: https://docs.zowe.org/stable/extend/extend-desktop/mvd-dataservices/#using-dataservices-with-rbac

c/authService.c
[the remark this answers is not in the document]
DivergentEuropeans: I'll add a comment here as well to make that more clear

c/authService.c
```diff
@@ -173,16 +196,223 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+    return rc; // When rbac isn't enabled, we don't try to check the auth query
+  }
+  rc = zisCheckEntity(privilegedServerName, userName, class, entity, access,
+      &reqStatus);
+  zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_DEBUG2,
+           "RBAC check occurred for entity '%s' class '%s' access '%d' , rc: %d", entity, class, access, rc);
+  return rc;
+}
+
+int getProfileNameFromRequest(char *profileName, StringList *parsedFile, char *method, int instanceID, HttpResponse *response) {
```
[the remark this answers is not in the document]
DivergentEuropeans: extractQuery seems to just loop through the query checking if StringListElt->next exists and setting entity and accessStr (because the query is already a SAF query) to use zisCheckEntity

getProfileNameFromRequest generates a query which we then use zisCheckEntity on. So extractQuery doesn't accomplish much because we already get entity and accessStr as part of the process of turning the URL into a SAF query

c/authService.c
```diff
@@ -173,16 +198,217 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+    if (pos >= profileNameBufSize) {
+      break;
+    }
+    pos += snprintf(profileName + pos, profileNameBufSize - pos, ".%s", subUrl[index]);
+    index++;
+  }
+  if (pos >= profileNameBufSize) {
+    char errMsg[256];
+    snprintf(errMsg, sizeof(errMsg), "Generated SAF query longer than %d\n", profileNameBufSize - 1);
+    zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING, errMsg);
```
ifakhrutdinov, line 404: Can you use the following instead of writing to `errMsg`?
```c
zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
        "Generated SAF query longer than %d\n", profileNameBufSize - 1);
```

c/authService.c
```diff
@@ -173,16 +198,217 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+    }
+    pos = snprintf(profileName, profileNameBufSize, "%s.%d.CFG.%s.%s.%s", productCode, instanceID, pluginID, method, scope); 
+  } else if (strcmp(type, "core") == 0) {
+    if (rootServiceName == NULL) {
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
+       "Broken SAF query. Missing root service name.\n");
+      return -1;
+    }
+    pos = snprintf(profileName, profileNameBufSize, "%s.%d.COR.%s.%s", productCode, instanceID, method, rootServiceName); 
+  } else if (pos < 0) {
```
ifakhrutdinov, line 387: Should this be just a standalone if-statement?

c/authService.c
```diff
@@ -173,16 +198,217 @@ static int serveAuthCheck(HttpService *service, HttpResponse *res) {
...
+     "Internal string encoding error.\n");
+    return -1;
+  }
+  // Child endpoints housed via subUrl
+  int index = 0;
+  while (index < SAF_SUB_URL_SIZE && strcmp(subUrl[index], "") != 0) {
+    if (pos >= profileNameBufSize) {
+      break;
+    }
+    pos += snprintf(profileName + pos, profileNameBufSize - pos, ".%s", subUrl[index]);
```
ifakhrutdinov, line 398: Do we want to also check if the result is < 0 since we're doing that for the other calls?

c/datasetService.c
```diff
@@ -188,6 +188,7 @@ void installDatasetContentsService(HttpServer *server) {
 
   HttpService *httpService = makeGeneratedService("datasetContents", "/datasetContents/**");
   httpService->authType = SERVICE_AUTH_NATIVE_WITH_SESSION_TOKEN;
+  httpService->authorizationType = SERVICE_AUTHORIZATION_TYPE_NONE;
```
ifakhrutdinov, line 191: Do 3rd party plugins have to do this as well and then rebuild the binaries, or will everything work without recompilation?

c/zss.c
```diff
@@ -319,10 +321,76 @@ static void setPrivilegedServerName(HttpServer *server, JsonObject *mvdSettings,
...
+  if (request->username == NULL) {
+    // username is required to perform RBAC check
+    // would a check for (service->authType != SERVICE_AUTH_NATIVE_WITH_SESSION_TOKEN) be better?
+    return TRUE;
+  }
+
+  RbacAuthorizationData *rbacData = userData;
+
+  char method[16];
+  snprintf(method, sizeof(method), "%s", request->method);
```
ifakhrutdinov, line 338: Should we validate the method length and not proceed if it's too long?

c/zss.c
```diff
@@ -319,10 +321,76 @@ static void setPrivilegedServerName(HttpServer *server, JsonObject *mvdSettings,
...
+    if (dataserviceAuth) {
+      rbacParm = jsonObjectGetBoolean(dataserviceAuth, "rbac");
+    }
+  } else {
+    rbacParm = jsonAsBoolean(rbacObj);
+  }
+  return rbacParm;
+}
+
+static int getZoweInstanceId() {
```
ifakhrutdinov, line 369: Declare functions that take no parameters using a `void` argument, for example:
```c
int foo(void);
```
Otherwise it's considered "unspecified number of parameters" by the standard.

c/zss.c
```diff
@@ -319,10 +321,76 @@ static void setPrivilegedServerName(HttpServer *server, JsonObject *mvdSettings,
...
+  return atoi(instance);
+}
+
+/* Future custom ZSS authorization handlers go here */
+static void registerAuthorizationHandlers(HttpServer *server, bool rbacEnabled) {
+  if (!rbacEnabled) {
+    return;
+  }
+  RbacAuthorizationData *rbacData = (RbacAuthorizationData*) safeMalloc(sizeof(*rbacData), "Rbac Authorization Data");
+  if (rbacData) {
```
ifakhrutdinov, line 383: Shouldn't we report an error and maybe terminate if this is NULL? If we silently do nothing here, ZSS is less protected and that's not going to be discovered.

h/authService.h
```diff
@@ -26,9 +26,37 @@
 #include "httpserver.h"
 #include "dataservice.h"
 
+#define SAF_CLASS "ZOWE"
```
ifakhrutdinov, line 29: Can you be more specific? `ZOWE_SAF_CLASS` would be clearer.

PR #287, opened 2021-04-19
Test for Handle files with unknown extension

lchudinov:
Test for https://github.com/zowe/zowe-common-c/pull/213


lchudinov:
Not needed anymore

PR #307, opened 2021-06-20
Some code cleanup

lchudinov:

### DIFF
--- a/c/authService.c
+++ b/c/authService.c
@@ -351,7 +351,7 @@ static int makeProfileName(
            "Broken SAF query. Missing method.");
     return -1;
   }
-  // char someString[STRING_BUFFER_SIZE] = { snprintf(*someString, STRING_BUFFER_SIZE, type) };
+  int pos = 0;
   if (strcmp(type, "service") == 0) {
     if (pluginID == NULL) {
       zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
@@ -363,7 +363,7 @@ static int makeProfileName(
        "Broken SAF query. Missing service name.");
       return -1;
     }
-    snprintf(profileName, ZOWE_PROFILE_NAME_LEN + 1, "%s.%d.SVC.%s.%s.%s", productCode, instanceID, pluginID, serviceName, method);
+    pos = snprintf(profileName, ZOWE_PROFILE_NAME_LEN + 1, "%s.%d.SVC.%s.%s.%s", productCode, instanceID, pluginID, serviceName, method);
   } else if (strcmp(type, "config") == 0) {
     if (pluginID == NULL) {
       zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING,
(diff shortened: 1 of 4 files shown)


c/authService.c
```diff
@@ -355,67 +341,63 @@ static const char* makeProfileName(
...
+  int pos = strlen(profileName);
   while (index < SAF_SUB_URL_SIZE && strcmp(subUrl[index], "") != 0) {
-    snprintf(profileName, STRING_BUFFER_SIZE, "%s.%s", profileName, subUrl[index]);
+    if (pos > ZOWE_PROFILE_NAME_LEN) {
+      char errMsg[256];
+      snprintf(errMsg, sizeof(errMsg), "Generated SAF query longer than %d", ZOWE_PROFILE_NAME_LEN);
+      zowelog(NULL, LOG_COMP_ID_SECURITY, ZOWE_LOG_WARNING, errMsg);
+      return -1;
+    }
+    pos += snprintf(profileName + pos, ZOWE_PROFILE_NAME_LEN + 1 - pos, ".%s", subUrl[index]);
```
DivergentEuropeans, line 397: This code doesn't seem to function as intended 100%

1) I tried with `"http://rs28.rocketsoftware.com:63097/server/agent/configconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfigconfig"` 
and it doesn't activate the "SAF query too long" case because it doesn't have a compatible URL segment to do the check before quitting

It works with,
`"http://rs28.rocketsoftware.com:63097/server/agent/agentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagentagent/config"`

2. According to @1000TurquoisePogs , desired thing to happen to be consistent with App server:

```
looks like its:
-construct name without url. if name is already over limit, then 403
which is very unlikely ✔️ 
-otherwise continue by... adding portions of the url up to the limit. ✔️ 
-at the limit, just stop adding to the name and process what you have ❌ 

What this means is that very long urls get treated as a subset of a shorter portion of their url 
which is highly likely to be secure, because people are very likely to be auth-protecting categories of urls, not very minute differences at the end of the url 
```

I crossed out things that work as intended, I believe. If we make `return -1` ->` return 0` , then it will be satisfied. Because currently this returns when profileName reaches 246 characters, but returns -1 so fails authorization regardless

DivergentEuropeans (changes requested):

PR #310, opened 2021-07-01
Rbac code cleanup

lchudinov:
* Revert back to integer authType
* Remove mistakenly added file


DivergentEuropeans:
This is a good change

PR #330, opened 2021-08-30
Limit entity check to default class only to prevent misuse by ZSS

ifakhrutdinov:
## Proposed changes

A recent PR showed that it's likely that a person who doesn't know the implications of exposing ZIS services may accidentally  create a security vulnerability. This PR closes the ability to pass a SAF class to the entity check service which will make it impossible to misuse it by any future ZSS changes. Additionally, a bug has been fixed where the access service would always return an error if a non empty class was provided.

## Type of change

## PR Checklist

## Testing
The standard tests against the security endpoints need to be performed.


lchudinov (approved):
Looks good to me. Thank you.

PR #347, opened 2021-09-21
Fix sso env var setting

1000TurquoisePogs:
This is a short potential bugfix that uses the same sso env var setting as seen in app-server.
The ZWED vars are essential for zss to get the config values it needs, but don't appear to be read in zss setup, but instead in app-server setup. So, that could break when app-server is in a container instead


1000TurquoisePogs:
I didnt realize that due to this line https://github.com/zowe/zss/blob/staging/bin/zssServer.sh#L31 this code is not needed.

PR #348, opened 2021-09-21
Add zlux-root-dir env var

1000TurquoisePogs:
This is a small fix to be able to resolve plugin definition relativeTo references that use the new value "ZLUX_ROOT_DIR" which is supposed to be the path to the zlux folders such as zlux-app-server. Here's an example of a plugin: https://github.com/zowe/zowe-install-packaging/blob/staging/files/zlux/config/plugins/org.zowe.configjs.json

### DIFF
--- a/bin/configure.sh
+++ b/bin/configure.sh
@@ -12,6 +12,9 @@
 # - ROOT_DIR
 # - WORKSPACE_DIR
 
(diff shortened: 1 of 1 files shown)


struga0258:
PR fixed the issue i was having.

PR #384, opened 2021-12-08
Unique cookie name for multi-instance zss use

1000TurquoisePogs:
Depends upon https://github.com/zowe/zowe-common-c/pull/262
Zowe v1 has an issue where you can't reach 2 unrelated ZSS from the same browser because they have identical cookie names and so one will overwrite the cookie value from the other, making only 1 valid at a time.
In order to fix this, something unique needs to be in the cookie name.
I implemented the same logic as seen in app-server which had the same problem and resolved it.
the cookie name has the suffix of the server port, unless HA is being used in which case the HA instance has a unique ID.
In both cases, the cookie name becomes unique enough that collision will no longer commonly happen.
This is a breaking change in that the cookie name has changed, so code that acts upon the cookie name will fail.


c/zss.c
```diff
@@ -1041,6 +1041,21 @@ static void readAgentAddressAndPort(JsonObject *serverConfig, JsonObject *envCon
   }
 }
 
+static char* generateCookieName(JsonObject *envConfig, int port) {
+  int cookieLength=256;
+  char *cookieName = safeMalloc(cookieLength+1, "CookieName");
+  char *zoweInstanceId = jsonObjectGetString(envConfig, "ZOWE_INSTANCE");
+  int haInstanceCount = jsonObjectGetNumber(envConfig, "ZWE_HA_INSTANCES_COUNT");
+  if (haInstanceCount > 1) {
```
lchudinov, line 1049: please check that `zoweInstanceId != NULL` as well

c/zss.c
```diff
@@ -1041,6 +1041,21 @@ static void readAgentAddressAndPort(JsonObject *serverConfig, JsonObject *envCon
   }
 }
 
+static char* generateCookieName(JsonObject *envConfig, int port) {
+  int cookieLength=256;
+  char *cookieName = safeMalloc(cookieLength+1, "CookieName");
+  char *zoweInstanceId = jsonObjectGetString(envConfig, "ZOWE_INSTANCE");
```
lchudinov, line 1047: `envConfig` contains only `ZWED`-prefixed environment variables. Shouldn't you get other env vars with `getenv()`?

lchudinov (changes requested):
The changes make sense to me, but please fix the way how you obtain env vars.

lchudinov (approved):
Now it looks good to me.

c/zss.c
```diff
@@ -1041,6 +1041,25 @@ static void readAgentAddressAndPort(JsonObject *serverConfig, JsonObject *envCon
   }
 }
 
+static char* generateCookieName(JsonObject *envConfig, int port) {
+  int cookieLength=256;
+  char *cookieName = safeMalloc(cookieLength+1, "CookieName");
+  char *zoweInstanceId = getenv("ZOWE_INSTANCE");
+  char *haInstanceCountStr = getenv("ZWE_HA_INSTANCES_COUNT");
```
ifakhrutdinov, line 1048: These calls can return `NULL`  if there are no such environmental variables.

c/zss.c
```diff
@@ -1041,6 +1041,25 @@ static void readAgentAddressAndPort(JsonObject *serverConfig, JsonObject *envCon
...
+  int haInstanceCount=0;
+  if (haInstanceCountStr != NULL) {
+    haInstanceCount = atoi(haInstanceCountStr);
+  }
+  if (haInstanceCount > 1 && zoweInstanceId != NULL) {
+    snprintf(cookieName, cookieLength, "%s.%s", SESSION_TOKEN_COOKIE_NAME, zoweInstanceId);
```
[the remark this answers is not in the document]
ifakhrutdinov: `snprintf` can fail, since this is related to security we might want to check it so the results of `generateCookieName` aren't unpredictable.

c/zss.c
```diff
@@ -1041,6 +1041,25 @@ static void readAgentAddressAndPort(JsonObject *serverConfig, JsonObject *envCon
   }
 }
 
+static char* generateCookieName(JsonObject *envConfig, int port) {
```
ifakhrutdinov, line 1044: Do we want to have a counterpart function for cookie removal?

c/zss.c
```diff
@@ -1041,6 +1041,25 @@ static void readAgentAddressAndPort(JsonObject *serverConfig, JsonObject *envCon
...
+  int cookieLength=256;
+  char *cookieName = safeMalloc(cookieLength+1, "CookieName");
```
ifakhrutdinov, line 1046: It's a good practice to check for `NULL`. Perhaps we should have a wrapper for `malloc`/`safeMalloc` which would abort if NULL is returned. We could use it for places where we don't want to continue execution if such a condition occurs.

lchudinov (changes requested):
Please fix all places where cookie name is hardcoded e.g.  https://github.com/zowe/zss/blob/791de0b1376663a320998b5188b8edb7c6136f10/c/zss.c#L274
and test your PR carefully.

PR #407, opened 2022-02-14
Change default IP binding when ATTLS enabled

lchudinov:
## Proposed changes

This PR changes default IP binding to `0.0.0.0` when ATTLS enabled.

## Type of change

## PR Checklist

### DIFF
--- a/bin/configure.sh
+++ b/bin/configure.sh
@@ -11,107 +11,27 @@
 # Required variables on shell:
 # - ZWE_zowe_runtimeDirectory
 # - ZWE_zowe_workspaceDirectory
-convert_v2_to_v1() {
-  while read old_name new_name; do
-    old_val=$(eval echo "\$${old_name}")
-    new_val=$(eval echo "\$${new_name}")
-    if [ -z "${old_val}" -a -n "${new_val}" ]; then
-      export "${old_name}=${new_val}"
-    fi
-  done <<EOF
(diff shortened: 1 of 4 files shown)


lchudinov:
> @lchudinov why do we have to bind to `0.0.0.0` when AT/TLS?

We want to make ZSS listen on `0.0.0.0` when it supports secure connections.

gadyck:
I don't think it is necessary to bind to INADDR_ANY on z/OS to use AT-TLS to create secure connections, but as long as you or the client are not sensitive to the host IP address that gets assigned then it should be a transparent change.  Protocols such as Radius are sensitive to the host IP address and from personal experience do not use INADDR_ANY when communicating with a Radius server.  The following link describes how INADDR_ANY is resolved on z/OS https://www.ibm.com/docs/en/zos/2.4.0?topic=selection-how-tcpip-selects-source-ip-address

gadyck:
And a final thought... Rocket test systems running under z/VM are normally defined with a single IPv4 adapter and IPv6 is disabled.  On these systems the same host IP address will always be used.  This means that any testing done on these systems may not expose all possible variations.  In the "Real World" it is normal and likely to have more than one adapter defined and IPv6 to be enabled.  And from (recent bad) experience severe problems using INADDR_ANY or IPv6 enabled sockets never seen during Rocket testing can suddenly start occurring.

1000TurquoisePogs (approved):

PR #399, opened 2022-01-17
PARMLIB member prefix based on the module name

ifakhrutdinov:
## Proposed changes
Since ZIS can run from any module name now, it should support using different PARMLIB member names. This PR changes ZIS to use the first 4 letters (3-letter element ID + the sub-component) of the module name as the PARMLIB member name prefix.

## Type of change
Please delete options that are not relevant.
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Change in a documentation
- [ ] Refactor the code 
- [ ] Chore, repository cleanup, updates the dependencies.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## PR Checklist
Please delete options that are not relevant.
- [x] If the changes in this PR are meant for the next release / mainline, this PR targets the "staging" branch.
- [x] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zlux/blob/master/CONTRIBUTING.md))
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] New and existing unit tests pass locally with my changes
- [ ] video or image is included if visual changes are made
- [ ] Relevant update to CHANGELOG.md
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works, or describe a test method below

## Testing
ZIS needs to be able to run using the default PARMLIB member name `ZWESIP00` and when the ZIS module has a different name (the default is `ZWESIS01`), it should be looking for the corresponding PARMLIB member name.

|Module name|PARMLIB member name|

|ZIS module name   | `MEM=` in JCL|PARMLIB member name|
|----------|-------------:|:-------------:|
| ZWESIS01 | 00 (default) | ZWESIP00 |
| ZWESIS01 | 01 | ZWESIP01 |
| FOOBIS01 | 00 (default) | FOOBIP00 |
| FOOBIS01 | 01 | FOOBIP01 |

PR #410, opened 2022-02-18
Port "PARMLIB member prefix based on the module name"

ifakhrutdinov:
Same as https://github.com/zowe/zss/pull/399 but for v2. The submodule change has been excluded since it's already present in v2.


lchudinov (approved):
Looks good to me.

PR #419, opened 2022-03-01
pc bit check and set

1000TurquoisePogs:
This PR depends upon https://github.com/zowe/zlux-app-server/pull/201 to call a plugin init script which will among other things check if zss plugins that exist have the PC bit set on services. If not, it will attempt to add it. This makes the config more bullet proof because if plugins are moved or were packaged without attributes, the pc bit attribute can be added at runtime.

### DIFF
--- a/bin/configure.sh
+++ b/bin/configure.sh
@@ -12,10 +12,26 @@
 # - ZWE_zowe_runtimeDirectory
 # - ZWE_zowe_workspaceDirectory
 
-cd ${ZWE_zowe_runtimeDirectory}/components/app-server/share/zlux-app-server/bin
-. ./convert-env.sh
+COMPONENT_HOME=${ZWE_zowe_runtimeDirectory}/components/zss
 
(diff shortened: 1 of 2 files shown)


bin/configure.sh
```diff
@@ -12,10 +12,26 @@
...
+  fi
+fi
+
+# Consume server infrastructure environment variables if possible
+if [ -n "$ZLUX_ROOT_DIR" ]; then
+  cd ${ZLUX_ROOT_DIR}/zlux-app-server/bin/init
+  . ../utils/convert-env.sh
+  NO_NODE=1
+  CHECK_PC_BIT=1
+  . ./plugins-init.sh $NO_NODE $CHECK_PC_BIT
```
1000TurquoisePogs, line 34: Add a code comment about how this is being called in one case and not the other because the other is a TODO for future packaging options that i'm unsure about... maybe just comment the whole other case out?

1000TurquoisePogs:
Let's do the zss bit check on zss itself too

1000TurquoisePogs:
I reviewed this with Tim today and I'm going to merge now and continue testing.

PR #447, opened 2022-04-11
Support for SAF IDT

achmelo:
continuation of https://github.com/zowe/zss/pull/403

## Proposed changes

This PR addresses Issue: [*Link to Github issue within https://github.com/zowe/zlux/issues* if any]

## Type of change
- [ ] Refactor the code 

## PR Checklist
- [ ] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zlux/blob/master/CONTRIBUTING.md))


achmelo:
@ifakhrutdinov could you please take a look?

h/zis/services/auth.h
```diff
@@ -91,6 +100,7 @@ int zisAuthServiceFunction(CrossMemoryServerGlobalArea *globalArea,
 #define RC_ZIS_AUTHSRV_CUSTOM_CLASS_NOT_ALLOWED   20
 #define RC_ZIS_AUTHSRV_INPUT_STRING_TOO_SHORT     21
 #define RC_ZIS_AUTHSRV_MAX_RC                     21
+#define RC_ZIS_AUTHSRV_BAD_SAF_SERVICE_VERSION    22
```
ifakhrutdinov, line 109: `RC_ZIS_AUTHSRV_BAD_SAF_SERVICE_VERSION` must be 21 and `RC_ZIS_AUTHSRV_MAX_RC` must become 22.

c/zis/services/auth.c
```diff
@@ -90,13 +61,81 @@ static int handleVerifyPassword(AuthServiceParmList *parmList,
...
-  if (idta != NULL) {
-    safeFree(idta, sizeof(IDTA));
-    idta = NULL;
-  }
   return rc;
 }
 
+static int handleGenerateToken(AuthServiceParmList *parmList,
+                                const CrossMemoryServerGlobalArea *globalArea) {
+       if(parmList->safIdtService.safIdtServiceVersion > 1) {
```
ifakhrutdinov: Don't use magic numbers, define the version in the header via `#define`.

c/zis/services/auth.c
```diff
@@ -90,13 +61,81 @@ static int handleVerifyPassword(AuthServiceParmList *parmList,
...
+            return RC_ZIS_AUTHSRV_BAD_SAF_SERVICE_VERSION;
+       }
+    ACEE *acee = NULL;
```
ifakhrutdinov: Formatting is incorrect. Please refer to the rest of this file to format appropriately.

c/zis/services/auth.c
```diff
@@ -90,13 +61,81 @@ static int handleVerifyPassword(AuthServiceParmList *parmList,
...
+      int safRC = 0, racfRC = 0, racfRsn = 0;
+      int deleteSAFRC = 0, deleteRACFRC = 0, deleteRACFRsn = 0;
+      int rc = RC_ZIS_AUTHSRV_OK;
+      IDTA *idta = NULL;
+      int options = VERIFY_CREATE;
+
+      if (parmList->options & ZIS_AUTH_SERVICE_PARMLIST_OPTION_GENERATE_IDT) {
+        idta = (IDTA *) safeMalloc31(sizeof(IDTA), "Idta structure");
```
ifakhrutdinov: NULL check is needed.

h/zis/services/auth.h
```diff
@@ -51,12 +60,14 @@ typedef struct AuthServiceParmList_tag {
 #define ZIS_AUTH_SERVICE_PARMLIST_FC_VERIFY_PASSWORD 0
 #define ZIS_AUTH_SERVICE_PARMLIST_FC_ENTITY_CHECK 1
 #define ZIS_AUTH_SERVICE_PARMLIST_FC_GET_ACCESS 2
+#define ZIS_AUTH_SERVICE_PARMLIST_FC_GENERATE_TOKEN 3
   char userIDNullTerm[ZIS_AUTH_SERVICE_USER_ID_MAX_LENGTH + 1];
   char passwordNullTerm[ZIS_AUTH_SERVICE_PASSWORD_MAX_LENGTH + 1];
   /* up to 8 characters: */
   char classNullTerm[ZIS_AUTH_SERVICE_CLASS_MAX_LENGTH + 1];
   char options;
 #define ZIS_AUTH_SERVICE_PARMLIST_OPTION_GENERATE_IDT 0x80
```
ifakhrutdinov: Should `ZIS_AUTH_SERVICE_PARMLIST_OPTION_GENERATE_ID` be deprecated and not be checked since there is a separate function code which already means that action? I'd leave a comment telling not to use 0x80 because there is already client code which might set it.

h/zis/services/auth.h
```diff
@@ -51,12 +60,14 @@ typedef struct AuthServiceParmList_tag {
...
+#define ZIS_AUTH_SERVICE_PARMLIST_OPTION_IDT_APPL 0x40
```
ifakhrutdinov: Does it make sense to keep IDT specific options (at least the new ones) in `SafIdtService`?

ifakhrutdinov (changes requested):
Please see my comments.

h/zis/services/auth.h
achmelo: shouldn't it be RC_ZIS_AUTHSRV_BAD_SAF_SERVICE_VERSION=22 and RC_ZIS_AUTHSRV_MAX_RC=22 ?

ifakhrutdinov: Yep, correct, my bad.

c/zis/services/auth.c
achmelo: added constant

c/zis/services/auth.c
achmelo: fixed according to the rest of the file

c/zis/services/auth.c
achmelo: added null check and new RC

h/zis/services/auth.h
achmelo: added comment with the text that this parameter is deprecated

h/zis/services/auth.h
achmelo: moved IDT options to SafIdtService struct

achmelo:
@ifakhrutdinov could you please review the latest updates if they are in line with your requests?

ifakhrutdinov:
@achmelo thanks, will do

c/zis/services/auth.c
```diff
@@ -66,74 +66,77 @@ static int handleVerifyPassword(AuthServiceParmList *parmList,
...
-    acee_deleted:
-
-      FILL_SAF_STATUS(&parmList->safStatus, safRC, racfRC, racfRsn);
-      CMS_DEBUG(globalArea, "handleVerifyPassword() done\n");
-      if (idta != NULL) {
-        safeFree(idta, sizeof(IDTA));
-        idta = NULL;
-      }
-      return rc;
+   if(parmList->safIdtService.safIdtServiceVersion < SAF_IDT_SERVICE_CURRENT_VERSION) {
```
ifakhrutdinov: This should be `>` I think. You want to support older clients.

h/zis/services/auth.h
```diff
@@ -45,10 +45,17 @@ typedef struct AbendInfo_tag {
 #define ZIS_AUTH_SERVICE_ENTITY_MAX_LENGTH      255
 #define ZIS_AUTH_SERVICE_APPL_MAX_LENGTH     8
 
+#define SAF_IDT_SERVICE_CURRENT_VERSION         1
```
ifakhrutdinov: Please add a `ZIS_` prefix.

h/zis/services/auth.h
```diff
@@ -66,8 +73,7 @@ typedef struct AuthServiceParmList_tag {
   /* up to 8 characters: */
   char classNullTerm[ZIS_AUTH_SERVICE_CLASS_MAX_LENGTH + 1];
   char options;
-#define ZIS_AUTH_SERVICE_PARMLIST_OPTION_GENERATE_IDT 0x80
-#define ZIS_AUTH_SERVICE_PARMLIST_OPTION_IDT_APPL 0x40
+
```
ifakhrutdinov: Define something like
```c
#define ZIS_AUTH_SERVICE_PARMLIST_OPTION_RESERVED 0x80
```
That way no one will reuse it by accident. I don't think it's completely safe to reuse it because there is code out there which already uses it.

c/zis/services/auth.c
```diff
@@ -66,74 +66,77 @@ static int handleVerifyPassword(AuthServiceParmList *parmList,
...
+   ACEE *acee = NULL;
+   int safRC = 0, racfRC = 0, racfRsn = 0;
+   int deleteSAFRC = 0, deleteRACFRC = 0, deleteRACFRsn = 0;
+   int rc = RC_ZIS_AUTHSRV_OK;
+   IDTA *idta = NULL;
+   int options = VERIFY_CREATE;
+
+
+  idta = (IDTA *) safeMalloc31(sizeof(IDTA), "Idta structure");
+  if(idta == NULL) {
```
ifakhrutdinov: Hmm, actually, can we just pass a stack variable? For example,
```c
IDTA idta = {
  .eyecatcher = "IDTA",
  .version = IDTA_VERSION_0001,
  .idtLen = parmList->safIdtService.safIdtLen,
  .options = VERIFY_GENERATE_IDT,
  .idtPropIn = IDTA_End_User_IDT,
};
. . .
safRC = safVerify6(options, parmList->userIDNullTerm, parmList->passwordNullTerm, &acee, &racfRC, &racfRsn, &idta);
```
And even if we don't use a stack variable, you can get rid of the NULL checks because you don't continue if `idta` is `NULL`.

c/zis/client.c
```diff
@@ -162,25 +163,51 @@ int zisGenerateOrValidateSafIdt(const CrossMemoryServerName *serverName,
...
     return RC_ZIS_SRVC_SERVICE_FAILED;
   }
-  memcpy((void *)parmList.safIdt, (void *)safIdt, strlen(safIdt));
+  memcpy((void *)parmList.safIdtService.safIdt, (void *)safIdt, strlen(safIdt));
 
   int rc = authRequest(serverName, &parmList, status);
 
-  if (parmList.safIdtLen > 0) {
+  if (parmList.safIdtService.safIdtLen > 0) {
     memset((void *)safIdt, 0, ZIS_AUTH_SERVICE_PARMLIST_SAFIDT_LENGTH + 1);
```
ifakhrutdinov: How do we know that `safIdt` is large enough for `ZIS_AUTH_SERVICE_PARMLIST_SAFIDT_LENGTH + 1` and for `parmList.safIdtService.safIdtLen`?

c/zis/client.c
```diff
@@ -162,25 +163,51 @@ int zisGenerateOrValidateSafIdt(const CrossMemoryServerName *serverName,
...
-    memcpy((void *)safIdt, (void *)parmList.safIdt, parmList.safIdtLen);
+    memcpy((void *)safIdt, (void *)parmList.safIdtService.safIdt, parmList.safIdtService.safIdtLen);
```
ifakhrutdinov: You can't cast `const` away, it's an input parameter, isn't it? Why do we write into it?

If it's an output arg it shouldn't be `const`.

c/zis/client.c
```diff
@@ -162,25 +163,51 @@ int zisGenerateOrValidateSafIdt(const CrossMemoryServerName *serverName,
   }
   strncpy(parmList.passwordNullTerm, password, sizeof(parmList.passwordNullTerm));
 
-  parmList.options |= ZIS_AUTH_SERVICE_PARMLIST_OPTION_GENERATE_IDT;
-  parmList.safIdtLen = strlen(safIdt);
+  if (appl != NULL && strlen(appl) != 0) {
```
ifakhrutdinov, line 167: Is `appl = NULL` a valid condition for this function? If yes, that should be documented.

ifakhrutdinov (changes requested):
@achmelo thanks for the updates, please see my comments.

c/zis/client.c
achmelo: safIdt is initialized  `char safIdt[ZIS_AUTH_SERVICE_PARMLIST_SAFIDT_LENGTH + 1] = {0};` so I think it is large enough for the first one. I was not able to find the maximum length of the saf ID, but do you think 2049 bytes is not enough for a token?

c/zis/client.c
achmelo: const removed

h/zis/services/auth.h
achmelo: macro added

h/zis/services/auth.h
achmelo: added prefix

c/zis/services/auth.c
achmelo: done

c/zis/client.c
achmelo: I've added the doc for appl parameter

c/zis/services/auth.c
achmelo: there are compilation errors when I tried this(ERROR CCN3022 ../../c/zis/services/auth.c:80    "eyecatcher" is not a member of "struct IDTA_tag".
ERROR CCN3022 ../../c/zis/services/auth.c:87    "options" is not a member of "struct IDTA_tag".) . Would it be ok if I just remove the null check then?

ifakhrutdinov: That was just an example, the field name is probably different.

c/zis/services/auth.c
```diff
@@ -75,7 +111,7 @@ static int handleVerifyPassword(AuthServiceParmList *parmList,
   }
 
   if (idta != NULL) {
```
ifakhrutdinov: `idta` cannot be `NULL`.

c/zis/client.c
```diff
@@ -162,25 +164,51 @@ int zisGenerateOrValidateSafIdt(const CrossMemoryServerName *serverName,
...
+    if (strlen(appl) >= sizeof (parmList.safIdtService.applNullTerm)) {
+      status->baseStatus.serviceRC = RC_ZIS_AUTHSRV_INPUT_STRING_TOO_LONG;
+      return RC_ZIS_SRVC_SERVICE_FAILED;
+    }
+    strcpy(parmList.safIdtService.applNullTerm, appl);
+
+    parmList.options |= ZIS_AUTH_SERVICE_PARMLIST_OPTION_IDT_APPL;
```
ifakhrutdinov: I thought you wanted to move this field to `SafIdtService`?

h/zis/services/auth.h
```diff
@@ -43,6 +43,22 @@ typedef struct AbendInfo_tag {
...
+#define ZIS_SAF_IDT_SERVICE_CURRENT_VERSION         1
+
+/* The #define for ZIS_AUTH_SERVICE_PARMLIST_OPTION_GENERATE_IDT has been deprecated, and will be removed
+   in the future.
+  */
+typedef struct SafIdtService_tag{
+  int safIdtServiceVersion;
+  int safIdtLen;
+  #define ZIS_AUTH_SERVICE_PARMLIST_SAFIDT_LENGTH (2 * IDTA_IDT_BUFFER_LEN_MIN)
+  #define ZIS_AUTH_SERVICE_PARMLIST_OPTION_GENERATE_IDT 0x80
```
ifakhrutdinov: This is compared with the `options` field in `AuthServiceParmList` and it has the same value as `ZIS_AUTH_SERVICE_PARMLIST_OPTION_RESERVED`.

Since you've moved your constants to `SafIdtService`, logically the option field you check should also be in `SafIdtService`.

Also, please change `ZIS_AUTH_SERVICE_PARMLIST_OPTION_xxxx` to `ZIS_AUTH_SERVICE_SAFIDT_OPTION_xxxx`

c/zis/client.c
ifakhrutdinov: So, if `parmList.safIdtService.safIdtLen` is larger than 2049, you'll get buffer overflow. Or maybe not, because SAF would fail first:

```c
   idtaVar.idtBufferPtr = parmList->safIdtService.safIdt;
   idtaVar.idtBufferLen = sizeof(parmList->safIdtService.safIdt);
   idtaVar.idtLen = parmList->safIdtService.safIdtLen;
```

c/zis/services/auth.c
```diff
@@ -35,32 +35,68 @@ static int handleVerifyPassword(AuthServiceParmList *parmList,
...
-    idta->idtLen = parmList->safIdtLen;
-    idta->idtPropIn = IDTA_End_User_IDT;
-    options |= VERIFY_GENERATE_IDT;
-  }
 
   CMS_DEBUG(globalArea, "handleVerifyPassword(): username = %s, password = %s\n",
       parmList->userIDNullTerm, "******");
-  if (idta == NULL) {
-    safRC = safVerify(VERIFY_CREATE, parmList->userIDNullTerm,
+
```
ifakhrutdinov, line 41: Since you've reverted changes in this function, it no longer support the IDT option, doesn't it? I think the right thing to do would be checking
```c
if (parmList->options & ZIS_AUTH_SERVICE_PARMLIST_OPTION_RESERVED) {
  return SOME_BAD_RC;
}
```
That way you'll make sure that a previous version of ZSS which relied on this would fail explicitly.

c/zis/client.c
```diff
@@ -162,25 +164,51 @@ int zisGenerateOrValidateSafIdt(const CrossMemoryServerName *serverName,
...
-  memcpy((void *)parmList.safIdt, (void *)safIdt, strlen(safIdt));
+  memcpy((void *)parmList.safIdtService.safIdt, (void *)safIdt, strlen(safIdt));
 
   int rc = authRequest(serverName, &parmList, status);
 
-  if (parmList.safIdtLen > 0) {
+  if (parmList.safIdtService.safIdtLen > 0) {
     memset((void *)safIdt, 0, ZIS_AUTH_SERVICE_PARMLIST_SAFIDT_LENGTH + 1);
-    memcpy((void *)safIdt, (void *)parmList.safIdt, parmList.safIdtLen);
+    memcpy((void *)safIdt, (void *)parmList.safIdtService.safIdt, parmList.safIdtService.safIdtLen);
```
ifakhrutdinov: Nit: you don't need to cast these to `void*`, and in the other call sites too.

ifakhrutdinov: Is the result SAF IDT a normal string without a null-terminator? Can `\0` be a valid character in a SAF IDT? If yes, how will the caller distinguish between that and the null-terminator?

I'm asking because the SAF call returns the length of the result, that is, it doesn't operate with C-strings.

c/zis/client.c
achmelo: when I was testing this on the internal system, the token was usually around 360 so I don't think the difference can be that big. What would you suggest as a solution for this? Increase the length?

c/zis/client.c
achmelo: I have tried to move it, but I have failed. The code did not compile with the error coming from an unrelated part of the code.

``` 
../../c/zis/client.c:
ERROR CCN3046 ../../deps/zowe-common-c/h/xml.h:119   Syntax error.
ERROR CCN3277 ../../deps/zowe-common-c/h/xml.h:162   Syntax error: possible missing ')' or ','?
ERROR CCN3282 ../../deps/zowe-common-c/h/xml.h:162   The type of the parameters must be specified in a prototype.
CCN0793(I) Compilation failed for file ./../../c/zis/client.c.  Object file not created.
FSUM3065 The COMPILE step ended with return code 12.
FSUM3017 Could not compile ../../c/zis/client.c. Correct the errors and try again.
```

h/zis/services/auth.h
achmelo: moved to SafIdtService and renamed all ZIS_AUTH_SERVICE_* related to safidt

c/zis/services/auth.c
achmelo: added check and return bad version rc

ifakhrutdinov: You need to check that **before** you do any SAF calls.

h/zis/services/auth.h
```diff
@@ -47,17 +47,18 @@ typedef struct AbendInfo_tag {
...
-  #define ZIS_AUTH_SERVICE_PARMLIST_SAFIDT_LENGTH (2 * IDTA_IDT_BUFFER_LEN_MIN)
-  #define ZIS_AUTH_SERVICE_PARMLIST_OPTION_GENERATE_IDT 0x80
-  #define ZIS_AUTH_SERVICE_PARMLIST_OPTION_IDT_APPL 0x40
-  char safIdt[ZIS_AUTH_SERVICE_PARMLIST_SAFIDT_LENGTH];
+  #define ZIS_AUTH_SERVICE_SAFIDT_LENGTH (2 * IDTA_IDT_BUFFER_LEN_MIN)
+  #define ZIS_AUTH_SERVICE_SAFIDT_OPTION_GENERATE_IDT 0x80
+  #define ZIS_AUTH_SERVICE_SAFIDT_OPTION_IDT_APPL 0x40
+  char safIdt[ZIS_AUTH_SERVICE_SAFIDT_LENGTH];
   char applNullTerm[ZIS_AUTH_SERVICE_APPL_MAX_LENGTH + 1];
+  char options;
```
ifakhrutdinov: Can you have it somewhere in the beginning of the struct, maybe after `safIdtLen`? 

And you can't just remove it from `AithServiceParmList` because it'll break the code. Replace it with `char _padding0[1];` instead.

c/zis/services/auth.c
achmelo: moved higher

h/zis/services/auth.h
achmelo: moved and replaced

h/zis/services/auth.h
```diff
@@ -43,6 +43,23 @@ typedef struct AbendInfo_tag {
 #define ZIS_AUTH_SERVICE_PASSWORD_MAX_LENGTH    100
 #define ZIS_AUTH_SERVICE_CLASS_MAX_LENGTH       8
 #define ZIS_AUTH_SERVICE_ENTITY_MAX_LENGTH      255
+#define ZIS_AUTH_SERVICE_APPL_MAX_LENGTH     8
+
+#define ZIS_SAF_IDT_SERVICE_CURRENT_VERSION         1
+
+/* The #define for ZIS_AUTH_SERVICE_SAFIDT_OPTION_GENERATE_IDT has been deprecated, and will be removed
```
ifakhrutdinov: Does it make sense to have this comment in `AuthServiceParmList` where `options` and `ZIS_AUTH_SERVICE_SAFIDT_OPTION_GENERATE_IDT` used to be?

c/zis/client.c
achmelo: I don't know, this is the only documentation I found https://www.ibm.com/docs/en/zos/2.4.0?topic=reference-activating-using-idta-parameter-in-racroute-requestverify

c/zis/client.c
ifakhrutdinov: Discussed this with Andrej, it was a misunderstand about moving to a header instead of to a different struct.

c/zis/client.c
ifakhrutdinov: From the page:
> Identity Token Formats
> The Identity Token Area (IDTA) is a flexible interface designed to support different types of Identity Tokens (IDT) going forward. Currently RACROUTE has support for an IDT in the format of a JSON Web Token (JWT). **In the future RACROUTE may support other IDT formats**.

Since you only use JWT according to the code:
```c
idtaVar.idtType = IDTA_JWT_IDT_Type;
```
I suggest updating the doc for these public functions telling that it only supports JWT. You will not be able to support anything else in the future because of the API you have here (relying on `\0`).

c/zis/client.c
```diff
@@ -162,25 +164,51 @@ int zisGenerateOrValidateSafIdt(const CrossMemoryServerName *serverName,
...
   }
-  memcpy((void *)parmList.safIdt, (void *)safIdt, strlen(safIdt));
+  memcpy((void *)parmList.safIdtService.safIdt, (void *)safIdt, strlen(safIdt));
 
   int rc = authRequest(serverName, &parmList, status);
 
-  if (parmList.safIdtLen > 0) {
-    memset((void *)safIdt, 0, ZIS_AUTH_SERVICE_PARMLIST_SAFIDT_LENGTH + 1);
-    memcpy((void *)safIdt, (void *)parmList.safIdt, parmList.safIdtLen);
+  if (parmList.safIdtService.safIdtLen > 0) {
```
ifakhrutdinov, line 187: Help me understand the logic. Why do we need this block of code? If SAF doesn't return anything to us except for the return codes, can we not do this? If so, can we just make `safIdt` an input parameter only and mark it as `const`?

h/zis/client.h
```diff
@@ -420,6 +420,14 @@ int zisCallNWMService(const CrossMemoryServerName *serverName,
 
 #define RC_ZIS_SRVC_NWM_BUFFER_NULL               (ZIS_MAX_GEN_SRVC_RC + 1)
 
+/**
+ * @brief This function generates the SAF ID token for @userName and @password is provided and @safIdt is empty.
```
ifakhrutdinov: * Does this line miss some words? Perhaps "**when** userName and password are provided and safIdt is empty"?
* Can you actually reference using `@`?
* Shouldn't there be `@param` in front of the parameters?
* What if userName, password and safIdt are all non empty?
* What does this function return?

Here's an example of valid doc:
```/**
 * @brief The function retrieves basic user profile information into
 * a user-provided buffer.
 *
 * @param callAuthInfo This structure should contain a space-padded user ID,
 * an ACEE, both or none. They're the RACF call authorization.
 * @param startUserID The first user ID used for iteration. Use NULL to iterate
 * from the beginning.
 * @param profilesToExtract The number of profiles to iterate over (extract).
 * @param result User-provided buffer for storing extracted profile info.
 * It must not be NULL.
 * @param profilesExtracted The number of profiles returned in @p result.
 * @param status This data structure is filled with status information in case
 * of a failure. It must not be NULL.
 *
 * @return One of the RC_RADMIN_xx return codes.
 */
```

h/zis/services/auth.h
achmelo: I thought it would make more sense to have a deprecated note where the parameter is. I have moved it to safservice.

c/zis/client.c
achmelo: thanks, I have updated the doc

achmelo: cast removed

c/zis/client.c
ifakhrutdinov: N/A

ifakhrutdinov (commented):
@achmelo, @JoeNemo, @1000TurquoisePogs this looks better. If someone else could have a look, it'd be great.

The PR needs to be squashed before merging and the DCO needs to be fixed. 

@struga0258 these changes must be tested very well as they're related to security and might potentially introduce security vulnerabilities.

ifakhrutdinov (commented):
@achmelo I think this also has conflicts which need to be resolved.

c/zis/client.c
```diff
@@ -138,17 +138,19 @@ static int authRequest(const CrossMemoryServerName *serverName,
 }
 
 /*
- *   safIdt - a buffer for SAF IDT token. The buffer must be ZIS_AUTH_SERVICE_PARMLIST_SAFIDT_LENGTH + 1 bytes long.
+ *   safIdt - a buffer for SAF IDT token. The buffer must be ZIS_AUTH_SERVICE_SAFIDT_LENGTH + 1 bytes long.
+ *   appl - application name to be included in the audience claim of SAF IDT. It is an optional parameter and can be null.
  */
-int zisGenerateOrValidateSafIdt(const CrossMemoryServerName *serverName,
+int zisGenerateOrValidateSafIdtWithAppl(const CrossMemoryServerName *serverName,
```
ifakhrutdinov, line 144: Do we really need to pass the password to generate or verify a SAF IDT value?

c/zis/client.c
```diff
@@ -138,17 +138,19 @@ static int authRequest(const CrossMemoryServerName *serverName,
 }
 
 /*
- *   safIdt - a buffer for SAF IDT token. The buffer must be ZIS_AUTH_SERVICE_PARMLIST_SAFIDT_LENGTH + 1 bytes long.
+ *   safIdt - a buffer for SAF IDT token. The buffer must be ZIS_AUTH_SERVICE_SAFIDT_LENGTH + 1 bytes long.
```
ifakhrutdinov, line 141: We're looking at this code with @JoeNemo now, and there is a question whether this could be dangerous when versions of the ZIS service change and it expects a larger buffer, which will cause an buffer overflow?

c/zis/client.c
achmelo: password is there for safverify https://github.com/zowe/zss/blob/3b52c8fa7295e0cd81dc01608f7683fe503f48a9/c/zis/services/auth.c#L47

c/zis/services/auth.c
```diff
@@ -35,32 +35,72 @@ static int handleVerifyPassword(AuthServiceParmList *parmList,
...
-    options |= VERIFY_GENERATE_IDT;
-  }
 
   CMS_DEBUG(globalArea, "handleVerifyPassword(): username = %s, password = %s\n",
       parmList->userIDNullTerm, "******");
-  if (idta == NULL) {
-    safRC = safVerify(VERIFY_CREATE, parmList->userIDNullTerm,
-      parmList->passwordNullTerm, &acee, &racfRC, &racfRsn);
+
+  if (parmList->safIdtService.options & ZIS_AUTH_SERVICE_SAFIDT_OPTION_RESERVED) {
```
ifakhrutdinov: You should check the value in the `_padding0` field against this now reserved value to detect older clients trying to use this new service in the old way, so that the call fails in an explicit manner.

c/zis/client.c
achmelo: potentially it could be an issue, although there is little chance. but this code is already present, so I don't think it will make much difference to change it in the future then changing it now

c/zis/services/auth.c
```diff
@@ -35,32 +35,72 @@ static int handleVerifyPassword(AuthServiceParmList *parmList,
...
+    parmList->passwordNullTerm, &acee, parmList->safIdtService.applNullTerm, &racfRC, &racfRsn, idta);
   } else {
     safRC = safVerify6(options, parmList->userIDNullTerm,
-      parmList->passwordNullTerm, &acee, &racfRC, &racfRsn, idta);
+    parmList->passwordNullTerm, &acee, &racfRC, &racfRsn, idta);
   }
+
   CMS_DEBUG(globalArea, "safVerify(VERIFY_CREATE) safStatus = %d, RACF RC = %d, "
       "RSN = %d, ACEE=0x%p\n", safRC, racfRC, racfRsn, acee);
   if (idta != NULL) {
```
ifakhrutdinov: This can't be `NULL`.

c/zis/client.c
ifakhrutdinov: So the RACF service cannot function without the user password?

ifakhrutdinov: We're trying to make sure this isn't opening up any way to use that in a bad way.

achmelo: how would racf verify the user if there is no password?

c/zis/services/auth.c
achmelo: done

c/zis/services/auth.c
achmelo: removed

ifakhrutdinov (approved):
The security expert explained that it's okay to host such a SAF call in our ZIS service and it's okay to generate tokens for users.

lgtm now

ifakhrutdinov:
@1000TurquoisePogs I think Andrej needs to fix DCO, can you please merge?

@achmelo @struga0258 can you please check if the previous version of ZSS SAF IDT service fail when used with the updated ZIS?

ifakhrutdinov:
And can you please squash the commits before merging?

c/zis/client.c
JoeNemo: A conversation of a SAF Developer indicated that we need to tighten the security of who/how this service is called.  But not in this PR.

JoeNemo (approved):
Approved.  But we need better security for callers of this service soon.

ifakhrutdinov:
@achmelo DCO should have been fixed and the commits should have been squashed. The PR checklist isn't filled out either.

PR #466, opened 2022-04-22
Fix jobname

1000TurquoisePogs:
## Proposed changes

Some users have reported that the jobname of some servers have changed. This appears to effect zss when the gateway group is running, but regardless a solution could be to force the zowe prefix to include the zowe instance. While modifying a user-configured variable to include another user-configured variable this seems wrong, it's what the rest of the code has been doing for a very long time, so I'm repeating it to be in alignment.

## Type of change

## PR Checklist
- [x] I have added tests that prove my fix is effective or that my feature works, or describe a test method below

## Testing

Start Zowe up via STC to see if the jobname of app-server is for example ZWE1DS1 versus ZWEDS1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to the ZSS package will be documented in this file.
 
 ## Recent Changes
 
+## v1.28.0
+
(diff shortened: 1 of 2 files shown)


1000TurquoisePogs:
Merged to be consistent with https://github.com/zowe/zlux-app-server/pull/225/files

PR #477, opened 2022-05-25
[bugfix] resolved `The module zssServer was not found` issue when LIBPATH is set 

aditya-ranshinge:
Signed-off-by: Aditya Ranshinge <aranshinge@rocketsoftware.com>

## Proposed changes
If LIBPATH is set and ztop/exporer-ip enabled  then zss fails with `CEE3501S The module zssServer was not found.` 
so added `.` value in LIBPATH env variable if not present already

## Type of change
- [ ] Refactor the code 

## PR Checklist
- [x] I have made corresponding changes to the documentation


DivergentEuropeans (approved):
This is a small change and it looks harmless. Also fixes an internal bug 👍

ifakhrutdinov:
Can you elaborate what exactly happens when " ztop/exporer-ip is enabled". Doe that change `ZSS_SCRIPT_DIR`? If so, adding the current directory to `LIBPATH` seems more like a workaround.

aditya-ranshinge:
@ifakhrutdinov By default `.` is present in `LIBPATH` env variable. If `LIBPATH` is set manually by overwriting existing values then it wont have `.`  in it. If `.` is not present and ztop or explorer-ip is enabled then zss throws `CEE3501S The module zssServer was not found.`  error. 
It doesn't change `ZSS_SCRIPT_DIR`.

ifakhrutdinov:
In that case perhaps it's better to add the actual absolute path to the zss binary to `LIBPATH` instead of `.` because the latter means different things once you change directories. `.` seems too "loose". Anyway, that's just my opinion. Happy to hear what the others think.

aditya-ranshinge:
@ifakhrutdinov  I agree with you, solution should be zss specific so added  zss dir path instead of `.`

ifakhrutdinov:
@adityadr-joy is there a point in having that complex `if` check? That is, can we always just add that path.

Also, the comment is now wrong. And it should probably say why we do that and not what we do.

aditya-ranshinge:
@ifakhrutdinov  if condition prevents redundant  addition of zss dir path to LIBPATH. If `.` or `ZSS_SCRIPT_DIR`  paths are present in LIBPATH then addition of `ZSS_SCRIPT_DIR` is not required

bin/zssServer.sh
```diff
@@ -13,6 +13,14 @@ export _BPXK_AUTOCVT=ON
 COMPONENT_HOME="${ZWE_zowe_runtimeDirectory}/components/zss"
 ZSS_SCRIPT_DIR="${COMPONENT_HOME}/bin"
 
+# this is to resolve LIBPATH current working dir issue. If LIBPATH is set manually by overwriting all existing values then cwd doesn't get resolved
+if [ -n "$LIBPATH" ]
+then
+  if [ "$(echo "${LIBPATH}" | grep ":.:")" != "${LIBPATH}" ] && [ "$(echo "${LIBPATH}" | cut -c1-2)" != ".:" ] && [ "$(echo "${LIBPATH}" | tail -c 3)" != ":." ] && [ "$(echo "${LIBPATH}" | grep "${ZSS_SCRIPT_DIR}")" != "${LIBPATH}" ];
```
ifakhrutdinov:
```sh
[ "$(echo "${LIBPATH}" | grep "${ZSS_SCRIPT_DIR}")" != "${LIBPATH}" ]
```
Wouldn't this also be true if `ZSS_SCRIPT_DIR` is a substring of a longer path? I.e. this would be a false positive.

bin/zssServer.sh
```diff
@@ -13,6 +13,14 @@ export _BPXK_AUTOCVT=ON
 COMPONENT_HOME="${ZWE_zowe_runtimeDirectory}/components/zss"
 ZSS_SCRIPT_DIR="${COMPONENT_HOME}/bin"
 
+# this is to resolve LIBPATH current working dir issue. If LIBPATH is set manually by overwriting all existing values then cwd doesn't get resolved
+if [ -n "$LIBPATH" ]
```
ifakhrutdinov: If `LIBPATH` is empty, shouldn't we also set it?

bin/zssServer.sh
ifakhrutdinov: I think this condition is super long and complicated. Let's say we always add that directory, will it cause issues? Do we do that multiple times?

@JoeNemo @1000TurquoisePogs what do you think?

bin/zssServer.sh
adityadr-joy: No errors if LIBPATH is empty

aditya-ranshinge: If LIBPATH is empty then no need to set it as nothing breaks when LIBPATH  is empty

bin/zssServer.sh
aditya-ranshinge: @ifakhrutdinov addition of ZSS_SCRIPT_DIR without if condition doesn't cause any issue. As it is run by script it doesn't get affected by restarts.

bin/zssServer.sh
ifakhrutdinov: Do we know why? Can that be system-dependent?

aditya-ranshinge: you are right, there is inconsistent behavior when LIBPATH is empty.

ifakhrutdinov:
I guess this looks fine. Can some bash guru confirm this? @1000TurquoisePogs ?

1000TurquoisePogs (changes requested):
Can you test the fix in one of two other ways:

1) Remove "export", keeping this env var change local
2) Keep "export", but move the logic to start.sh instead of zssServer.sh

A concern was raised about LIBPATH being a global variable and  if zssServer.sh is called by a human (it is sometimes) then the export can pollute the active shell.
start.sh should never be called by a human, so its fine there. And, if not an 'export', i think it's also file in zssServer.sh
So, please try one of these two and see if it still resolves your issue.
Putting marking the PR as needs work until these are tried.

aditya-ranshinge:
> 

@1000TurquoisePogs both suggested ways resolves this issue. First looks more optimal so removed export to keep this var change local .

1000TurquoisePogs (approved):

PR #488, opened 2022-07-18
login bug for mock server fixed by changing cookie name

TaoYeNa:
## Proposed changes

Fixing the bug which causes user with mock server not able to login to zowe desktop by adding port number to the cookie name which ZSS MOCK server should send.
**This PR addresses Issue:** 
 User not able to login into the zowe desktop with the Mock Server. Server returns code 401.

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/bin/zssServer.sh
+++ b/bin/zssServer.sh
@@ -13,6 +13,9 @@ export _BPXK_AUTOCVT=ON
 COMPONENT_HOME="${ZWE_zowe_runtimeDirectory}/components/zss"
 ZSS_SCRIPT_DIR="${COMPONENT_HOME}/bin"
 
+# this is to resolve ZSS bin path in LIBPATH variable.
+LIBPATH="${LIBPATH}:${ZSS_SCRIPT_DIR}"
+
 # this is to resolve (builtin) plugins that use ZLUX_ROOT_DIR as a relative path. if it doesnt exist, the plugins shouldn't either, so no problem
 if [ -z "${ZLUX_ROOT_DIR}" ]; then
   if [ -d "${ZWE_zowe_runtimeDirectory}/components/app-server/share" ]; then


mock/server.py
```diff
@@ -165,13 +169,14 @@ def genVolserId():
...
 @app.route('/login', methods=['POST'])
+
 def login():
     if request.get_json()['username'] == global_username:
         if request.get_json()['password'] == global_password:
             if global_password_expired == False:
                 resp = make_response("Login Successful")
-                resp.set_cookie("jedHTTPSession", "xrQdqvc2J7WWlew2OXU1RtwwUXXD9KGJ35x5IZTrSrX18y80OVBI8A")
-                return resp
+                resp.set_cookie("jedHTTPSession." + global_port, "xrQdqvc2J7WWlew2OXU1RtwwUXXD9KGJ35x5IZTrSrX18y80OVBI8A")
```
1000TurquoisePogs: Putting a port here works in some cases, but not the default case.
A better default is to use the env var "ZWE_zowe_cookieIdentifier" which defaults to "1" in a real zowe install.
https://github.com/zowe/zowe-install-packaging/blob/v2.x/master/example-zowe.yaml#L301=

TaoYeNa: So in real-time, the cookie name would be starting with jedHTTPSession.1? But this could cause an error with the mock server since the mock server needs a specific port number after the dot "."  .

1000TurquoisePogs: yes, it would be  jedHTTPSession.1
Here's how it works in ZSS: https://github.com/zowe/zss/blob/v2.x/staging/c/zss.c#L1171-L1175
The app-server uses this same logic, so basically the app-server may be expecting "jedHTTPSession.1" rather than "jedHTTPSession.5000" depending upon environment vars & config files.
You can still set the mock server to use port 5000, I'm just saying that in a real zowe install, zss wouldn't be presenting "jedHTTPSession.7557" (7557 is the default port), it would be presenting "jedHTTPSession.1"

TaoYeNa: got it. Should I change it to .1 instead in server.py?

DivergentEuropeans: > yes, it would be jedHTTPSession.1
Here's how it works in ZSS: https://github.com/zowe/zss/blob/v2.x/staging/c/zss.c#L1171-L1175
The app-server uses this same logic, so basically the app-server may be expecting "jedHTTPSession.1" rather than "jedHTTPSession.5000" depending upon environment vars & config files.
You can still set the mock server to use port 5000, I'm just saying that in a real zowe install, zss wouldn't be presenting "jedHTTPSession.7557" (7557 is the default port), it would be presenting "jedHTTPSession.1"

The ".1" case I believe is the HA case. For non-HA mode, cookie is the ZSS port. 
from zssHandler.js in sso-auth
`const zssPort = serverConf.agent.https && serverConf.agent.https.port ? serverConf.agent.https.port : serverConf.agent.http.port;
    this.zssCookieName = zluxUtil.isHaMode() ? COOKIE_NAME_BASE + zoweInstanceId : COOKIE_NAME_BASE + zssPort;`

DivergentEuropeans: That's why when we use mock server, if used in the regular development app server environment case, would be in this way - otherwise default mock server case this doesn't successfully log in

DivergentEuropeans:
```
const isHaMode = module.exports.isHaMode = function () {
  const haInstanceCount = +process.env['ZWE_HA_INSTANCES_COUNT'];
  return (!isNaN(haInstanceCount) && haInstanceCount > 1);
}
```

we can check for HA case, ok

1000TurquoisePogs:
I realized there's a bug in v2.0 to v2.2 app-server currently expects a different cookie name than zss provides.
But then in v2.3 staging zss was fixed.
https://github.com/zowe/zss/blob/v2.x/staging/c/zss.c#L1158-L1172 is the logic to follow for the mock server.

mock/server.py
```diff
@@ -7,22 +7,30 @@
...
+
 ###  TODO:
 #RBAC routes for permissions
 #Permissions changed to array [7,7,7] to better represent all the different types of permissoins and add permission check function
 
 global_username = "mock"
 global_password = "pass"
 global_password_expired = False
-
+#set port number that server uses,could be change to any number
```
DivergentEuropeans: While I agree with the isHaMode change (and at a glance, it looks like it will work 👍 ) this comment isn't actually true. We can set the port number yes, but I don't think the server will respect it, because we set the variable but we don't actually tell the server to use the variable. Most likely from 
https://stackoverflow.com/questions/20212894/how-do-i-get-flask-to-run-on-port-80
you will want to change
`if __name__ == '__main__':
    app.run(debug=True)`
    (last 2 lines of server.py) to something like 
    (maybe w/ small variation, have to test)
    `if __name__ == '__main__':
    app.run(port=global_port, debug=True)`

mock/server.py
```diff
@@ -7,22 +7,30 @@
...
+os.environ['FLASK_RUN_PORT'] = '5000'
```
DivergentEuropeans: We don't want to set it to 5000, instead we want to default it to 5000 only if the user env variable isn't specified (because we want the server to respect an env varialbe) so instead of
`os.environ['FLASK_RUN_PORT'] = '5000'`
`...`
`global_port = os.getenv('FLASK_RUN_PORT')`

we want something like
`global_port = os.getenv('FLASK_RUN_PORT') if os.getenv('FLASK_RUN_PORT') != undefined else '5000'`

DivergentEuropeans (changes requested):

DivergentEuropeans (approved):

DivergentEuropeans:
Changes were addressed and code respects env variables

PR #491, opened 2022-07-21
Allow faster zss startup via JS usage

1000TurquoisePogs:
By using the same plugin registration logic that app-server uses, conditional to app-server not being present, the zss server can startup quicker in a standard zowe environment. This saves about 1 minute of startup time.
The only issue with this PR is that it is straight code duplication. I copied a file from zlux-app-server into here.
I do not expect this file to need many changes over time, so I figure the performance tradeoff is worthwhile.


JoeNemo (approved):
This looks good.

PR #513, opened 2022-09-21
Fix jobname

1000TurquoisePogs:
In 2.3 the jobname of zss changed from something like ZWE1SZ to just "SZ" because the env var it used to get from app-server is now undefined. There is a correct env var to use, and should have been pointed at directly but seems to have been missed.


JoeNemo (approved):
Looks Good.

PR #525, opened 2022-10-13
feat: map distinguished user to RACF ID 

achmelo:
## Proposed changes

Add the possibility to pass a distinguished user ID for user mapping the same way as it is possible with the client certificate. 
By providing a user ID obtained from the distributed authorization server(e.g. Okta, keycloak,..), IRRSIM00 will return the RACF ID that is associated with it. 

## Type of change
- [ ] Refactor the code 

## PR Checklist
- [ ] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zlux/blob/master/CONTRIBUTING.md))


c/certificateService.c
```diff
@@ -210,9 +254,9 @@ static int serveMappingService(HttpService *service, HttpResponse *response)
   return 0;
 }
 
-void installCertificateService(HttpServer *server)
+void installUserMappingService(HttpServer *server)
 {
-  HttpService *httpService = makeGeneratedService("CertificateService", "/certificate/x509/**");
+  HttpService *httpService = makeGeneratedService("UserMappingService", "/user-mapping/**");
```
1000TurquoisePogs: This looks like a breaking change - you can't rename URLs like this without a v3 because someone might be using it.

1000TurquoisePogs (changes requested):
I think this should be reviewed by @ifakhrutdinov or @JoeNemo but until then, can you add a description and reconsider the URL change?

c/certificateService.c
achmelo: this is something I would like to discuss. Is it better to keep the misleading URL or rather do this change if it is probable that API ML is the only user of this service.

c/certificateService.c
```diff
@@ -118,36 +119,80 @@ static void respondWithBadRequest(HttpResponse *response) {
...
   if (!strcmp(request->method, methodPOST))
   {
-    RUsermapParamList *userMapCertificateStructure 
-      = (RUsermapParamList*)safeMalloc31(sizeof(RUsermapParamList),"RUsermapParamList");
-    memset(userMapCertificateStructure, 0, sizeof(RUsermapParamList));
-
-    if(request->contentLength > sizeof(userMapCertificateStructure->certificate) || request->contentLength < 0) {
-      respondWithBadRequest(response);
-      return 0;
+    RUsermapParamList userMapStructure;
```
ifakhrutdinov: For the future, you can do this:
```c
RUsermapParamList userMapStructure = {0};
```
More concise, less error-prone.

c/certificateService.c
```diff
@@ -118,36 +119,80 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+
+    memset(&userMapStructure, 0, sizeof(RUsermapParamList));
+
+    char translatedURL[strlen(request->uri) +1];
```
ifakhrutdinov: Should we limit this length? It doesn't feel right to allow users control a buffer size in memory via something they pass via web.

c/certificateService.c
```diff
@@ -118,36 +119,80 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+    memcpy(translatedURL, request->uri, strlen(request->uri));
+    a2e(translatedURL, sizeof(translatedURL) - 1);
+    char *found = strstr(translatedURL,"x509");
```
ifakhrutdinov: `translatedURL` is not null-terminated. What could happen is,
* Someone sends an empty/incorrect request
* We still find `x509` further down the road, because someone else has used one before
* We use it for the wrong caller

c/certificateService.c
```diff
@@ -118,36 +119,80 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+        }
+
+        userMapStructure.certificateLength = request->contentLength;
+        memset(userMapStructure.certificate, 0, request->contentLength);
+        memcpy(userMapStructure.certificate, request->contentBody, request->contentLength);
+
+        userMapStructure.functionCode = MAP_CERTIFICATE_TO_USERNAME;
+    } else {
+    //    Distinguished ID to user mapping
+        char *nb = copyStringToNative(request->slh, request->contentBody, strlen(request->contentBody));
```
ifakhrutdinov: * What does `nb` mean?
* What does `errB` mean?

c/certificateService.c
```diff
@@ -118,36 +119,80 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+
+        char *distinguishedId = jsonObjectGetString(jsonObject, "dn");
+        if(distinguishedId == NULL || strlen(distinguishedId) == 0) {
+            respondWithBadRequest(response, "dn field not included in request body");
+            return 0;
+        } else if(strlen(distinguishedId) > sizeof(userMapStructure.distinguishedName)){
+            respondWithBadRequest(response, "The length of the distinguished name is more than 246 bytes");
+            return 0;
+        }
+        e2a(distinguishedId, strlen(distinguishedId));
```
ifakhrutdinov: I think it's better to `e2a` the result buffer, not the source buffer. Who knows how else it's used, you don't control it.

ifakhrutdinov (changes requested):
Please see my comments.

c/certificateService.c
achmelo: I have added `= {0};` as suggested

c/certificateService.c
achmelo: is it ok if we return an error message in case of URI longer than 64? Or better to limit it to less?

c/certificateService.c
achmelo: sorry, I think I made a mistake here, now I removed `-1`

c/certificateService.c
achmelo: provided longer names for variables

c/certificateService.c
achmelo: converting the result buffer now

achmelo:
> Please see my comments.

Thank you for review, I have made the changes, please take a look.

achmelo:
@ifakhrutdinov please, did you have a chance to review my updates?

c/certificateService.c
ifakhrutdinov: Yes, any user input must be validated and if it doesn't conform to the API requirements, an error is returned.

c/certificateService.c
```diff
@@ -133,13 +133,17 @@ static int serveMappingService(HttpService *service, HttpResponse *response)
...
-    char translatedURL[strlen(request->uri) +1];
+    int urlLength = strlen(request->uri);
+    if(urlLength < 0 || urlLength > 64) {
+        respondWithBadRequest(response, "URL length is not in range from 0 to 64 bytes.");
+        return 0;
+    }
+    char translatedURL[urlLength];
     memcpy(translatedURL, request->uri, strlen(request->uri));
-    a2e(translatedURL, sizeof(translatedURL) - 1);
+    a2e(translatedURL, sizeof(translatedURL));
```
ifakhrutdinov, line 108: Wouldn't this still leave the array unterminated?

Did you mean?
```c
char translatedURL[urlLength + 1];
```

c/certificateService.c
```diff
@@ -169,11 +173,11 @@ static int serveMappingService(HttpService *service, HttpResponse *response)
             respondWithBadRequest(response, "The length of the distinguished name is more than 246 bytes");
             return 0;
         }
-        e2a(distinguishedId, strlen(distinguishedId));
+
         userMapStructure.distinguishedNameLength = strlen(distinguishedId);
         memset(userMapStructure.distinguishedName, 0, strlen(distinguishedId));
         memcpy(userMapStructure.distinguishedName, distinguishedId, strlen(distinguishedId));
-
+        e2a(userMapStructure.distinguishedName, strlen(userMapStructure.distinguishedName));
```
ifakhrutdinov: Will `strlen` even work here? The buffer doesn't seem to be null-terminated. Why not use `userMapStructure.distinguishedNameLength`?

c/certificateService.c
```diff
@@ -118,36 +119,85 @@ static void respondWithBadRequest(HttpResponse *response) {
...
   {
-    RUsermapParamList *userMapCertificateStructure 
-      = (RUsermapParamList*)safeMalloc31(sizeof(RUsermapParamList),"RUsermapParamList");
-    memset(userMapCertificateStructure, 0, sizeof(RUsermapParamList));
+    RUsermapParamList userMapStructure = {0};
 
-    if(request->contentLength > sizeof(userMapCertificateStructure->certificate) || request->contentLength < 0) {
-      respondWithBadRequest(response);
-      return 0;
+    memset(&userMapStructure, 0, sizeof(RUsermapParamList));
```
ifakhrutdinov: You don't need this line.

c/certificateService.c
```diff
@@ -118,36 +119,85 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+        return 0;
+    }
+    char translatedURL[urlLength];
+    memcpy(translatedURL, request->uri, strlen(request->uri));
+    a2e(translatedURL, sizeof(translatedURL));
+    char *found = strstr(translatedURL,"x509");
+
+    if(found != NULL) {
+    //  Certificate to user mapping
+        if(request->contentLength > sizeof(userMapStructure.certificate) || request->contentLength < 0) {
```
ifakhrutdinov: What if the content length is 0?

c/certificateService.c
```diff
@@ -118,36 +119,85 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+          respondWithBadRequest(response, "The length of the certificate is longer than 4096 bytes");
+          return 0;
+        }
+
+        userMapStructure.certificateLength = request->contentLength;
+        memset(userMapStructure.certificate, 0, request->contentLength);
+        memcpy(userMapStructure.certificate, request->contentBody, request->contentLength);
```
ifakhrutdinov: Is `contentBody` `NULL` if the `contextLength` is 0?

c/certificateService.c
```diff
@@ -118,36 +119,85 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+
+        userMapStructure.functionCode = MAP_CERTIFICATE_TO_USERNAME;
+    } else {
+    //    Distinguished ID to user mapping
+        char *bodyNativeEncoding = copyStringToNative(request->slh, request->contentBody, strlen(request->contentBody));
```
ifakhrutdinov: Is the content body guaranteed to be null-terminated? Should we use the content length instead?

c/certificateService.c
```diff
@@ -118,36 +119,85 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+        char errorBuffer[2048];
+        Json *body = jsonParseUnterminatedString(request->slh, bodyNativeEncoding, strlen(bodyNativeEncoding), errorBuffer, sizeof(errorBuffer));
```
ifakhrutdinov: Even the function name says `UnterminatedString` and at the same time `strlen` is used.

c/certificateService.c
```diff
@@ -118,36 +119,85 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+        JsonObject *jsonObject = jsonAsObject(body);
+
+        char *distinguishedId = jsonObjectGetString(jsonObject, "dn");
+        if(distinguishedId == NULL || strlen(distinguishedId) == 0) {
```
ifakhrutdinov: Please be consistent with blanks around keywords, brackets, operands, etc.

c/certificateService.c
```diff
@@ -118,36 +119,85 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+            return 0;
+        } else if(strlen(registry) > sizeof(userMapStructure.registryName)){
+             respondWithBadRequest(response, "The length of the registry name is more than 255 bytes");
+             return 0;
+         }
+
+        userMapStructure.registryNameLength = strlen(registry);
+        memset(userMapStructure.registryName, 0, strlen(registry));
+        memcpy(userMapStructure.registryName, registry, strlen(registry));
+        e2a(userMapStructure.registryName, strlen(userMapStructure.registryName));
```
ifakhrutdinov: Again, I don't believe that `registryName` is guaranteed to be null-terminated.

ifakhrutdinov (changes requested):
Please see my comments.

Martin-Zeithaml (commented):
1) The null-terminated problem with ```distinguishedName``` and ```registryName``` seems to be solved.
2) Breaking-change needs to be discussed and PR mark as "Breaking change"

c/certificateService.c
```diff
@@ -118,36 +119,83 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+    RUsermapParamList userMapStructure = {0};
 
-    if(request->contentLength > sizeof(userMapCertificateStructure->certificate) || request->contentLength < 0) {
-      respondWithBadRequest(response);
+    int urlLength = strlen(request->uri);
+    if (urlLength < 0 || urlLength > 64) {
+      respondWithBadRequest(response, "URL length is not in range from 0 to 64 bytes.");
       return 0;
     }
+    char translatedURL[urlLength + 1];
```
ifakhrutdinov, line 106: Bug: this is still left unterminated.

You can do
```c
char translatedURL[urlLength + 1] = {0};
```

Or use `strcpy` instead of `memcpy`.

c/certificateService.c
```diff
@@ -109,7 +110,7 @@ static void respondWithInvalidMethod(HttpResponse *response) {
     finishResponse(response);
 }
 
-static void respondWithBadRequest(HttpResponse *response) {
+static void respondWithBadRequest(HttpResponse *response, char *value) {
```
ifakhrutdinov: What is `value`? Is it the error message?

c/certificateService.c
```diff
@@ -118,36 +119,83 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+      if (request->contentLength > sizeof(userMapStructure.certificate) || request->contentLength < 1) {
+        respondWithBadRequest(response, "The length of the certificate is longer than 4096 bytes");
+        return 0;
+      }
 
-    userMapCertificateStructure->certificateLength = request->contentLength;
-    memset(userMapCertificateStructure->certificate, 0, request->contentLength);
-    memcpy(userMapCertificateStructure->certificate, request->contentBody, request->contentLength);
+        userMapStructure.certificateLength = request->contentLength;
+        memset(userMapStructure.certificate, 0, request->contentLength);
```
ifakhrutdinov: Nit: this line does nothing:
* `userMapStructure` is 0 anyway because
```c
RUsermapParamList userMapStructure = {0};
```
* `userMapStructure.certificate` is immediately overwritten by `request->contentBody` with the same length (`request->contentLength`) as in the `memset` call

c/certificateService.c
```diff
@@ -118,36 +119,83 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+        memcpy(userMapStructure.certificate, request->contentBody, request->contentLength);
+
+        userMapStructure.functionCode = MAP_CERTIFICATE_TO_USERNAME;
+    } else {
+    //    Distinguished ID to user mapping
+        char *bodyNativeEncoding = copyStringToNative(request->slh, request->contentBody, request->contentLength);
+        char errorBuffer[2048];
+        Json *body = jsonParseUnterminatedString(request->slh, bodyNativeEncoding, request->contentLength, errorBuffer, sizeof(errorBuffer));
```
ifakhrutdinov, line 128: This can be `NULL` if a bad JSON.

c/certificateService.c
```diff
@@ -118,36 +119,83 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+        JsonObject *jsonObject = jsonAsObject(body);
```
ifakhrutdinov, line 133: Also can be `NULL`.

c/certificateService.c
```diff
@@ -118,36 +119,83 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+        if (distinguishedId == NULL || strlen(distinguishedId) == 0) {
+          respondWithBadRequest(response, "dn field not included in request body");
+          return 0;
+        } else if (strlen(distinguishedId) > sizeof(userMapStructure.distinguishedName)) {
+          respondWithBadRequest(response, "The length of the distinguished name is more than 246 bytes");
+          return 0;
+        }
+
+        userMapStructure.distinguishedNameLength = strlen(distinguishedId);
+        memset(userMapStructure.distinguishedName, 0, strlen(distinguishedId));
```
ifakhrutdinov: This line seems to be useless because we overwrite the same amount of bytes on the next line. What did you intend here?

c/certificateService.c
```diff
@@ -118,36 +119,83 @@ static void respondWithBadRequest(HttpResponse *response) {
...
+        if (registry == NULL || strlen(registry) == 0) {
+          respondWithBadRequest(response, "registry field not included in request body");
+          return 0;
+        } else if (strlen(registry) > sizeof(userMapStructure.registryName)) {
+          respondWithBadRequest(response, "The length of the registry name is more than 255 bytes");
+          return 0;
+        }
+
+        userMapStructure.registryNameLength = strlen(registry);
+        memset(userMapStructure.registryName, 0, strlen(registry));
```
ifakhrutdinov: Same here.

ifakhrutdinov (changes requested):
Please see my comments. There is at least one bug related to null-termination, some issues with error handling of JSON and minor nits related to code that does nothing (or perhaps it did intend something else?).

c/certificateService.c
```diff
@@ -173,7 +179,6 @@ static int serveMappingService(HttpService *service, HttpResponse *response)
         }
 
         userMapStructure.distinguishedNameLength = strlen(distinguishedId);
-        memset(userMapStructure.distinguishedName, 0, strlen(distinguishedId));
```
Martin-Zeithaml: I believe, the intent was to clear entire field with zeroes and then copy the field, which is redundant because of next line and fact, the null terminator is not used by the macro, which relies on the length fields only. Removing the line is OK.

c/certificateService.c
```diff
@@ -118,36 +119,87 @@ static void respondWithBadRequest(HttpResponse *response) {
...
 {
   HttpRequest *request = response->request;
 
-  if (!strcmp(request->method, methodPOST))
-  {
-    RUsermapParamList *userMapCertificateStructure 
-      = (RUsermapParamList*)safeMalloc31(sizeof(RUsermapParamList),"RUsermapParamList");
-    memset(userMapCertificateStructure, 0, sizeof(RUsermapParamList));
+  if (!strcmp(request->method, methodPOST)) {
+    RUsermapParamList userMapStructure = {0};
```
JoeNemo: Why is safeMalloc31 being removed?  This is making the service even less 64-bit ready.  Shouldn't all this be moved to use ALLOC_STRUCT31

JoeNemo:
I propose that we merge the really small change from PR 546 into this PR because it is going to need to change the exact same places, and perhaps in a confusing way.   The change in PR 546 makes the IRRSIM00 linkage correct for 64-bit, but does not address the incorrect stack allocation.  There is a new comment in the review of certificateService.c to point out where things are wrong.

c/certificateService.c
ifakhrutdinov: There is an advantage: you don't have to manage this memory; considering how many exits points there are and the current design, that would have been error prone. For 64-bit this doesn't work of course.

I suggest the following structure:

```c
static int serveMappingService(HttpService *service, HttpResponse *response) {
  ALLOC_STRUCT31(...);

  if (error) {
    respondWithBadRequest(...);
    goto out;
  }
 
  if (anotherError) {
    respondWithBadRequest(...);
    goto out;
  } 
  
out:
  FREE_STRUCT31(...);
  return 0;
}

That way it's easier to reason what gets released when, and harder to miss a "free" call.

Comment (ifakhrutdinov) [reply c/certificateService.c]: `serveMappingService` could also be refactored not to have to have so many exits points, e.g. by splitting into separate functions (parm validation, RACF parm initialization, the RACF call itself).

Comment (achmelo) [reply c/certificateService.c]: can we make these changes in another PR? e.g. PR546.

**Comment by achmelo** [comment]
> I propose that we merge the really small change from PR 546 into this PR because it is going to need to change the exact same places, and perhaps in a confusing way. The change in PR 546 makes the IRRSIM00 linkage correct for 64-bit, but does not address the incorrect stack allocation. There is a new comment in the review of certificateService.c to point out where things are wrong.

If we merge PR546 and revert the allocation change, would it fix 64-bit version?

File: c/certificateService.c (hunk 24 above)
Comment (achmelo) [reply c/certificateService.c]: I am asking because 10.1. is code freeze, we can either make these changes into upcoming release or leave it for the next release, but the bug in the 64-bit version would still be there. Do you think it will decrease the quality of the code?

File: c/certificateService.c [hunk 25]
```diff
@@ -110,159 +111,153 @@ static void respondWithInvalidMethod(HttpResponse *response) {
...
 	     These offsets are stable, and this avoids linker/pragma mojo */
-	  " LA 15,X'10' \n"
-          " LG 15,X'220'(,15) \n" /* CSRTABLE */
-          " LG 15,X'28'(,15) \n" /* Some RACF Routin Vector */
-          " LG 15,X'A0'(,15) \n" /* IRRSIM00 itself */
+	        " LLGT 15,X'10' \n"
+          " LLGT 15,X'220'(,15) \n" /* CSRTABLE */
+          " LLGT 15,X'28'(,15) \n" /* Some RACF Routin Vector */
+          " LLGT 15,X'A0'(,15) \n" /* IRRSIM00 itself */
 	  " LG 1,%0 \n"
```
Martin-Zeithaml: Based on the listing, calling of parameter list looks good:
```
* * * * *   S T O R A G E   O F F S E T   L I S T I N G   * * * * *
parmlist31          136-0:131       Class = automatic,           Location = 56(r9),                            Length = 8
```
Code to be executed:
```
  LG       r6,parmlist31(,r9,56)     * From ASM_PREFIX
  LLGT ...                           * 4 x LLGT
  LG       r1,0(r6,)                 * Parameter list
  SAM31
  BALR 14,15
  SAM64
```

achmelo:
replaced by PR https://github.com/zowe/zss/pull/556

PR #553, opened 2023-01-26
Fix typo on schema of logger

1000TurquoisePogs:
"_zss.restdataset" could not be used because it was typod as _zss.resetdataset


JoeNemo (approved):
Approved.

PR #555, opened 2023-02-02
use proj version during build

1000TurquoisePogs:
ZSS was using a 'version.txt' to read its version number rather than zss.proj.env
I want to make our build technology more standard across repositories. So, since proj.env files are used in zowe-common-c, lets use them here too.
Likewise manifest.yaml contains a real version in the zlux plugins, so to have reusable github actions, setting it as a real number here too will allow it to be incremented in the pipeline.


JoeNemo (approved):
Approved.

PR #556, opened 2023-02-03
map distributed user to mainframe ID

achmelo:
Signed-off-by: achmelo <a.chmelo@gmail.com>

## Proposed changes

original PR created from forked repo(closed due to missing permissions to run workflows): https://github.com/zowe/zss/pull/525

This PR addresses Issue: [*Link to Github issue within https://github.com/zowe/zss/issues* if any]

## Type of change
- [ ] Refactor the code 

## PR Checklist
- [ ] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zss/blob/v2.x/master/CONTRIBUTING.md))


c/certificateService.c
```diff
@@ -109,110 +91,86 @@ static void respondWithInvalidMethod(HttpResponse *response) {
...
   {
-    RUsermapParamList *userMapCertificateStructure 
-      = (RUsermapParamList*)safeMalloc31(sizeof(RUsermapParamList),"RUsermapParamList");
-    memset(userMapCertificateStructure, 0, sizeof(RUsermapParamList));
 
-    if(request->contentLength > sizeof(userMapCertificateStructure->certificate) || request->contentLength < 0) {
-      respondWithBadRequest(response);
-      return 0;
+    int urlLength = strlen(request->uri);
+    if(urlLength < 0 || urlLength > 17) {
```
ifakhrutdinov: Why 17? I suggest you define this value (i.e. via a `#define`).

c/certificateService.c
```diff
@@ -109,110 +91,86 @@ static void respondWithInvalidMethod(HttpResponse *response) {
...
     }
 
-    userMapCertificateStructure->certificateLength = request->contentLength;
-    memset(userMapCertificateStructure->certificate, 0, request->contentLength);
-    memcpy(userMapCertificateStructure->certificate, request->contentBody, request->contentLength);
+    char translatedURL[urlLength + 1];
+    strcpy(translatedURL, request->uri);
+    a2e(translatedURL, sizeof(translatedURL));
+    char *x509URI = strstr(translatedURL, "x509");
+    char *dnURI = strstr(translatedURL, "dn");
```
ifakhrutdinov: What does "dn" stand for? Distinguished name?

c/certificateService.c
```diff
@@ -109,110 +91,86 @@ static void respondWithInvalidMethod(HttpResponse *response) {
...
-        &userMapCertificateStructure->useridLengthRacf,
-        &userMapCertificateStructure->certificateLength,
-        &userMapCertificateStructure->applicationIdLength,
-        &userMapCertificateStructure->distinguishedNameLength,
-        &userMapCertificateStructure->registryNameLength
-    );
-#endif
+    if(x509URI != NULL) {
+    //  Certificate to user mapping
+        if(request->contentLength > 4096 || request->contentLength < 1) {
```
ifakhrutdinov: Is 4096 an arbitrary limit?

c/certificateService.c
```diff
@@ -109,110 +91,86 @@ static void respondWithInvalidMethod(HttpResponse *response) {
...
+        if (body == NULL) {
+            respondWithJsonError(response, "JSON in request body has incorrect structure.", 400, "Bad Request");
+            return 0;
+        }
+        JsonObject *jsonObject = jsonAsObject(body);
+        if (jsonObject == NULL) {
+            respondWithJsonError(response, "Object in request body is not a JSON type.", 400, "Bad Request");
+            return 0;
+        }
+        char *distinguishedId = jsonObjectGetString(jsonObject, "dn");
```
ifakhrutdinov, line 139: This can be `NULL`.

c/certificateService.c
achmelo: changed to macro

c/certificateService.c
achmelo: yes, dn stands for distinguished name, there is full name now

c/certificateService.c
achmelo: it is maximum length of the certificate from R_usermap docs, since there is this check as a part of the rusermap service in common-c I have removed it.

c/certificateService.c
achmelo: I added null check for distinguishedId and registry

c/certificateService.c
ifakhrutdinov: Ok, but FYI the function in zowe-common-c doesn't check negative values which are allowed by the type.

achmelo: I can add the check for negative values, I was not aware that it would be possible

ifakhrutdinov (approved):
Ok, this looks reasonable.

@struga0258 , we should test this service thoroughly, especially with bad parameters. For example, a missing certificate, a certificate that is too large, etc.

@achmelo, though the function you call should verify some of its args, it's always good to verify parameters provided by users, because you can never trust them. So, I would personally keep the certificate length check, but you'd need to know what a correct certificate is. One way to do that would be having a separate type for certificates and put it in the header, e.g.
```c
typedef struct Certificate_tag {
  unsigned length;
  char data[MAX_CERTIFICATE_LENGTH];
} Certificate;
```
The type name might be different.

ifakhrutdinov:
@1000TurquoisePogs please add someone else to the review list, perhaps someone familiar with this service. Or if you wish, you can also review it. It's good to have another set of eyes on this.

1000TurquoisePogs (approved):
I'm not as much of an expert here, but I also reviewed this and it looks like correct use of the previously merged zowe-common-c code.

PR #557, opened 2023-02-07
Add ZIS dynamic build plugin into build and package

1000TurquoisePogs:

## Proposed changes
This adds the building of the zis dynamic plugin (via build_dynamic.sh) to the zss build, and puts its output into the zss pax. This is meant to be a plugin that sits by the side of ZIS so that other zis plugins can call it for accessing zowe-common-c utilities dynamically rather than statically compiled within.

## Type of change

## PR Checklist

## Testing
@ifakhrutdinov do we have examples of how to use this plugin? we should make some and could even be a nice test.


1000TurquoisePogs:
the only user action might be to update their zwesip00 to reference this plugin when needed.
otherwise, this should be a painless upgrade.

JoeNemo (approved):
Approved.

PR #560, opened 2023-02-21
Update ZIS message.h with naming and typo fixes

1000TurquoisePogs:
While looking through message.h for ZIS i spotted many occurrences of old names (not ZIS) and a few things that looked like typos, so some or all of these suggestions could be merged hopefully.

### DIFF
--- a/h/zis/message.h
+++ b/h/zis/message.h
@@ -305,7 +305,7 @@
 #define ZISDYN_LOG_DEV_MODE_MSG                 ZISDYN_LOG_DEV_MODE_MSG_ID" "ZISDYN_LOG_DEV_MODE_MSG_TEXT
 
 #define ZISDYN_LOG_BAD_ZIS_VERSION_MSG_ID       ZIS_MSG_PRFX"0214E"
-#define ZISDYN_LOG_BAD_ZIS_VERSION_MSG_TEXT     "Bad cross-memory server version: expected [%d.%d.%d, %d.0.0], found %d.%d.%d"
+#define ZISDYN_LOG_BAD_ZIS_VERSION_MSG_TEXT     "Bad cross-memory server version: expected [%d.%d.%d, %d.0.0), found %d.%d.%d"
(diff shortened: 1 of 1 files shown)


ifakhrutdinov:
I believe the use of "Zowe Cross-Memory Server" was intentional. ZIS is a code name. When we discussed the name originally we chose something (i.e. a full name) that sysprogs would be happy with.

ifakhrutdinov:
ZSS to Zowe is a good change in my opinion. I'm open to having ZIS somewhere to clarify the relation (maybe an additional message?), we can think about that.

h/zis/message.h
```diff
@@ -305,7 +305,7 @@
 #define ZISDYN_LOG_DEV_MODE_MSG                 ZISDYN_LOG_DEV_MODE_MSG_ID" "ZISDYN_LOG_DEV_MODE_MSG_TEXT
 
 #define ZISDYN_LOG_BAD_ZIS_VERSION_MSG_ID       ZIS_MSG_PRFX"0214E"
-#define ZISDYN_LOG_BAD_ZIS_VERSION_MSG_TEXT     "Bad cross-memory server version: expected [%d.%d.%d, %d.0.0), found %d.%d.%d"
+#define ZISDYN_LOG_BAD_ZIS_VERSION_MSG_TEXT     "Bad cross-memory server version: expected [%d.%d.%d, %d.0.0], found %d.%d.%d"
```
ifakhrutdinov: This is incorrect. `)` was there intentionally.

ifakhrutdinov (changes requested):

h/zis/message.h
ifakhrutdinov: It's a half-open interval. For example, [2.1.3, 3.0.0) would say that we support the versions from 2.1.3 to, but not including, 3.0.0.

1000TurquoisePogs: ohhh. ok.

PR #565, opened 2023-02-27
Consolidate shell logic to reduce overhead

1000TurquoisePogs:
When looking at the process tree i noticed making start.sh run zssServer.sh is counted as 2 processes. This separation isnt really needed with what we are doing, so I inlined all of zssSever.sh content into start.sh and made zssServer.sh point to start.sh for compatibility. This should give us one less process.


JoeNemo (approved):
Simple code move.  Approved

PR #570, opened 2023-03-10
Remove some dead logging code that was rewritten in v2

1000TurquoisePogs:
I was trying to debug something and needed the loggers turned up, but got stuck looking at unused logging code that has since been replaced. So, let's remove this to prevent confusion, because it was already commented out.


JoeNemo (approved):
Approved

PR #574, opened 2023-03-20
Add description about loglevel naming

1000TurquoisePogs:
Note about periods in logLevel strings to avoid confusion about jq syntax


JoeNemo (approved):
Approved.

Issue #576, opened 2023-03-22
Support various keyring types in ZSS


Martin-Zeithaml:
**Is your feature or enhancement request related to a problem or limitation? Please describe**

Resolve the https://github.com/zowe/api-layer/issues/2686 and https://github.com/zowe/community/issues/1851 for zss.

**Describe your enhancement idea**

Support various keyring types in ZSS.

**Describe alternatives you've considered**

None.

**Provide any additional context**

None.

1000TurquoisePogs:
Note #581 as a possible solution which needs testing by someone with such keyrings.
Also note that #581 is simply a YAML edit, meaning, you don't even need this PR to have such support. If the keyrings work, then this means older versions of ZSS could use such keyrings just by use of multi-yaml-config which overrides the defaults.

PR #581, opened 2023-03-30
Update defaults.yaml for other keyring types

1000TurquoisePogs:
This PR makes the logic on keyring detection & use more generic, so that other types of keyrings could be loaded.
This does not test whether ZSS can handle such keyrings, only that ZSS will be provided them.
For https://github.com/zowe/zss/issues/576


JoeNemo (approved):
We may get some new bugs if we bring in key rings that we didn't see before, but we have to go in this direction.

PR #582, opened 2023-03-31
space info metadata

1000TurquoisePogs:
## Proposed changes
Updates the /datasetMetadata api for space, prime, and secnd fields for telling size information about each dataset.

Getting the space for "prime" correct has been challenging. DSCB extent info on how many tracks/cylinders are occupied can be way different from what something like ispf reports, depending on several factors such as unit  type.
So, this api currently sometimes uses the cyl/track info found, and sometimes instead relies upon a combination of other size-related fields found in the DSCB. This compromise passes most of my testing, with remaining areas of improvement when size type=block or blocksize is small.

## Type of change

## PR Checklist

## Testing
You can test this in the zowe editor, point it at any datasets you want.
I left some warnings in the code in case datasets i didn't find in my own environment were discovered, so that logic can be tuned to accomodate them.

Bug: datasets that are block-sized dont have proper reporting of primary/secondary unit, not sure how the math works here yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to the ZSS package will be documented in this file.
 
 ## Recent Changes
 
+## `2.8.0`
+
+- Enhancement: /datasetMetadata now returns primarySize, secondarySize, and spaceUnit fields
+
 ## `2.7.0`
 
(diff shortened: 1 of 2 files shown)


JoeNemo (commented):
I think this still has debug printf's in it.

JoeNemo (commented):
I believe, now that I am remembering some more about storage apps, that some fields in the Format-1 DSCB are rarely used "raw" from the program, but need some interpretation by standard OS routines.

JoeNemo (approved):
Approved.  It will probably need some tweaking for block-sized allocations in the near future.

PR #583, opened 2023-04-05
Fix branch for github workflow

1000TurquoisePogs:
Switches pointer from users/jstruga/bump-zis to v2.x/main


JoeNemo (approved):
Approved.

PR #601, opened 2023-05-17
Add the ZIS dynamic linkage base plugin's version to env

ifakhrutdinov:
#### Overview
The version of the dynamic linkage plugin must be incremented every time it's shipped because it's located in LPA and, if not  one, ZIS will keep the current version in LPA thinking that the plugin has not changed (ZIS checks plugin versions when deciding  whether to refresh the LPA module).

## Proposed changes

This PR addresses Issue: [*Link to Github issue within https://github.com/zowe/zss/issues* if any]

This PR depends upon the following PRs:

## Type of change
Please delete options that are not relevant.
- [x] Bug fix (non-breaking change which fixes an issue)

## PR Checklist
Please delete options that are not relevant.
- [x] If the changes in this PR are meant for the next release / mainline, this PR targets the "staging" branch.
- [x] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zss/blob/v2.x/master/CONTRIBUTING.md))
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] New and existing unit tests pass locally with my changes
- [ ] video or image is included if visual changes are made
- [x] Relevant update to CHANGELOG.md
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works, or describe a test method below

### Steps to reproduce
* Start ZIS version v2.8.0 with the dynamic linkage plugin enabled
* Stop ZIS
* Downgrade that ZIS to version v2.7.0
* Start ZIS with the dynamic linkage plugin enabled again
* You should get the `ZWES0714E` error and ZIS will terminate

### Behavior after the fix
* Start ZIS version v2.x.0 with the dynamic linkage plugin enabled
* Stop ZIS
* Downgrade that ZIS to version v2.x-1.0
* Start ZIS with the dynamic linkage plugin enabled again
* You should get the `ZWES0018W` message for the dynamic linkage plugin and ZIS will **not** terminate

PR #602, opened 2023-05-19
Disable dynamic zis plugin by default

1000TurquoisePogs:
The dynamic zis plugin is likely to be used by extensions of zowe, but isnt needed by zowe's base.
To be efficient to system resources, it's better to leave this disabled and have products enable it when needed.


1000TurquoisePogs:
#601

PR #603, opened 2023-05-24
Adding symbol resolution for volser during obtaindscb1

1000TurquoisePogs:
This PR leverages the `resolveSymbol` function in zowe-common-c to resolve volsers that start with & during dscb lookup.
Previously, libraries such as in sys1 would fail in the zowe editor because when we went to lookup properties like record length, we failed because the volser was a symbol that needed to be resolved to do the lookup.
This fixes that, by resolving any volser string that comes into the dscb1 function if it starts with a &.


1000TurquoisePogs:
Examples of such symbols are
&SYSD1, &SYSR1
These come from IGCSI00, and the volser field there can only be 6 chars, so yes, the symbols in there would be max 6 too.

JoeNemo (approved):
Approved.

PR #633, opened 2023-07-27
Avoid using submodules because its annoying

1000TurquoisePogs:
It had been discussed that because we have a way of linking to different github places without use of submodules, we should just stop using submodules for zowe-common-c.
submodules are annoying to everyone, so why not use our proj files instead?


ifakhrutdinov:
Although I don't agree with the idea, you can already use submodules to track branches:

Setting things up:
```
git submodule set-branch --branch v2.x/staging -- zowe-common-c
git commit .gitmodule
```

Using it:
```
git clone --recursive git@github.com:zowe/zss.git
git submodule update --remote --recursive
```
This will bring the submodules to the latest state of their branches.

Briefly, I can see the following issues with both the approaches though:
* Not tracking concrete commits is error prone as there will be cases where something goes out of sync
* You cannot go back and check the state of the ZSS repo at a certain point because you wouldn't know what state of zowe-common-c it was pointing at that moment
* Merging, for example, `v2.x/staging` into `master` will get complicated as you'll have either conflicting `.gitmodules` or your proj files since they need to point to their own respective branches.

I know there is an attitude of "who cares if staging breaks", but with this approach the amount of issues like that will likely increase. And breaking staging is not good either in my opinion. We have complicated authorized code that can potentially bring the system down because someone will update zowe-common-c without taking into account the state of the ZSS repo.

Could explain more how submodules make your development process more complicated and how are they annoying? Perhaps I can share tips and tricks that can make the issues go away.

Let's discuss this.

PR #646, opened 2023-09-18
support for tls1.3

jordanfilteau1995:
Making ciphers and keyshares configurable.


JoeNemo:
@jordanfilteau1995 are you ready to merge?

jordanfilteau1995:
@JoeNemo I need Sean to sign off on leaving the "string" based cipher override for now because I don't have time to do the "english" to "openssl cipher enum" converter and global cipher override.

jordanfilteau1995:
I made it so you _can_ change if needed, but it shouldn't. There aren't many supported ciphers for TLSv1.3.

1000TurquoisePogs (approved):
I tested this directly (GET /plugins) and indirectly (login to desktop with sso and access the editor) and saw both that tls 1.3 was being used, and that i saw no regressions.

PR #663, opened 2023-11-02
Update zwes1606w message for troubleshooting

1000TurquoisePogs:
When you have a JWK error, you currently get output like this:

```
error initing session: 17
failed to obtain JWK, status = 9
ZWES1606W Failed to get JWK - failed to send HTTP request, retry in 10 seconds
```
9 = "failed to send http request", so this is redundant.
17 is mysterious.
and overall, this message spams when nothing is wrong but zowe isnt fully loaded yet.

this PR decreases the ZWES1606W message to only fire every 10 intervals (100 seconds by default), and skips the first 10 to potentially avoid 1606w entirely during normal timing events.
this PR eliminates the top 2 lines, consolidating them into ZWES1606W, which now will print:

```
ZWES1606W Failed to get JWK. rc=failed to send HTTP request (9), rsn=TLS error (17). Retry in 10 seconds
```


timgerstel (approved):

JoeNemo (approved):
Approved

PR #666, opened 2023-11-20
Add cipher customization support

1000TurquoisePogs:
adds reading array of iana cipher suite strings to customize ciphers

this can be tested by setting either zowe.network.server.tls.ciphers or components.zss.zowe.network.server.tls.ciphers to an array of IANA cipher strings.

this PR also externalizes the default ciphers into defaults.yaml so people can see what they are, though its a little hard to read.

when _zss.mvdserver and _zss.httpserver log levels are set to debug (3) or higher, you will see printout about the behavior - which ciphers were requested and which maps were found, and what the final result is.

to test, i set up my yaml like this:
```
  network:                                                                                                   
    server:                                                                                                  
      listenAddresses:                                                                                       
      - 0.0.0.0                                                                                              
      tls:                                                                                                   
        minTls: TLSv1.2                                                                                      
        maxTls: TLSv1.3                                                                                      
#        ciphers:                                                                                            
#        - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256                                                             
#        - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384                                                             
#        - TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256                                                     
#        - TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256                                                       
#        - TLS_CHACHA20_POLY1305_SHA256
```

by uncommenting or commenting, and switching maxTls to 1.2 or 1.3, I was able to check the site details in firefox to see if the right ciphers were being used and at what tls level
image

my testing matched the config & debug output.


c/zss.c
```diff
@@ -1162,13 +1164,36 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
...
-   *
+   * Takes a string of ciphers.
    * ciphers: 13021303003500380039002F00320033
    */
+    ciphers = jsonObjectGetString(httpsConfigObject, "ciphers");
+    zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG, "Cipher string override to %s\n", ciphers);
+  } else {
+    JsonArray *cipherArray = jsonObjectGetArray(httpsConfigObject, "ciphers");
+    int count = jsonArrayGetCount(cipherArray);
+    ciphers = (char *)safeMalloc((4*count)+1, "cipher list");
```
jordanfilteau1995: What does the four represent? Do you mean size of a `char *` which is four? If that's the case, just do `sizeof(char *)` for clarity as opposed to `4`.

c/zss.c
```diff
@@ -1162,13 +1164,36 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
...
+    for (int i = 0; i < count; i++) {
+      char *ianaName = jsonArrayGetString(cipherArray, i);
+      zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG, "Cipher request=%s\n", ianaName);
+      CipherMap *cipher = (CipherMap *)ianaCipherMap;
+      while (cipher->suiteId != NULL) {
```
jordanfilteau1995, line 1189: This is a nitpick, but it might be better to calculate the total number of elements and use that for a loop as opposed to relying on your static array ending with a null. It could also be the max string length of the cipher identifiers, I suppose, and if that's the case, please make a `define`.

c/zss.c
```diff
@@ -1162,13 +1164,36 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
...
+        if (!strcmp(ianaName, cipher->name)) {
```
jordanfilteau1995, line 1190: It might be nice to print out the ciphers that were "ignored" if we aren't validating the entirety of the list of the schema level.

jordanfilteau1995 (approved):
I don't have any major issues.

c/zss.c
1000TurquoisePogs: that might be a bug, let me fix.
ciphers are 4 chars long.

c/zss.c
1000TurquoisePogs: that makes sense

c/zss.c
1000TurquoisePogs: sure

PR #671, opened 2023-11-28
Add json logging

1000TurquoisePogs:
JSON logging was hidden behind a DEBUG flag. Instead it is easier to use zowelog in the form of "_zss.json"

Depends upon https://github.com/zowe/zowe-common-c/pull/418


JoeNemo (approved):
Approved.

JoeNemo (approved):
Approved.

Issue #599, opened 2023-05-11
High Availability mode session cookie name


pablocarle:
**Describe the bug**

Linked to https://github.com/zowe/zlux/issues/969

The function to resolve the session cookie name uses environment variables that are not available in Zowe V2 installer.
https://github.com/zowe/zss/blob/d0db646c0b579e43bdaddeda8ef4d854a86139df/c/zss.c#L1097
`ZOWE_INSTANCE` and `ZWE_HA_INSTANCES_COUNT`

This can result in errors in session refresh feature of zlux when started in HA mode.

**Steps to Reproduce**

1. Open Zowe Desktop in an HA-enabled Zowe instance

**Expected behavior**

`jedHTTPSession` should either take values of `jedHTTPSession.<port>` or `jedHTTPSession.<instance id>` depending on the Zowe mode.

**Describe your environment**

- Zowe version number (Check the Desktop login screen, or manifest.json in the Zowe install folder): nightly v2
- Install method (pax, smpe, kubernetes, github clone): pax
- Operating system (z/OS, kubernetes, etc) and OS version: z/OS

1000TurquoisePogs:
This was merged for v2.15.0

PR #675, opened 2023-12-11
Fix cookie name setup

1000TurquoisePogs:
ZSS was determining whether or not to use a cookie with the suffix of its http port, or with the suffix of the HA instance name, depending on whether it detected an environment variable that appears to never have been in v2, and maybe was from v1.

Now, I try to detect if we're in an HA mode depending upon the quantity of elements in the "haInstances" section of the zowe YAML.

This should be sufficient, since if you are not using HA, that section should be commented out.


1000TurquoisePogs:
Fix for https://github.com/zowe/zlux/issues/969

1000TurquoisePogs:
Fix for https://github.com/zowe/zss/issues/599

timgerstel (approved):

PR #687, opened 2024-02-14
Default zss to 64 bit for v3

1000TurquoisePogs:
Change zss startup to default to running zss 64 bit in all cases other than components.zss.agent.use64bit = "false"


JoeNemo (approved):
Approved.   This corrects a careless mistake from 2015.

PR #694, opened 2024-03-12
updating pointer for zowe-common-c.

jordanfilteau1995:
See: https://github.com/zowe/zowe-common-c/pull/438


JoeNemo (approved):
Approved.

PR #710, opened 2024-05-28
Set native tls off if attls is on

1000TurquoisePogs:
`tls: true` in our config controls whether this server uses http or https on the listening port.
While it has been defaulting to true unconditionally for a long time, it is better to default it to the newly introduced parameter of whether or not attls is enabled.
This way, it is easier to enable attls for zss.


JoeNemo (approved):
THe power of scripting!!

PR #715, opened 2024-06-14
Apply HEAPPOOLS64(OFF) to plugins-init

1000TurquoisePogs:
HEAPPOOLS64(OFF) was added in launcher and zowe-install-packaging to capture most occurrences of configmgr which requires this.
However, it was forgotten in this one place of zss, so I have added this PR.


JoeNemo (approved):
Approved.

Issue #713, opened 2024-06-11
Support curve customization


1000TurquoisePogs:
zowe.network.server.tls.curves is an array a user can set in zowe.yaml to customize crypto curves.
But, zss doesnt do anything with that array currently.

GSK handles curve customization much like cipher customization, which we already do here https://github.com/zowe/zss/blob/f44ce8a3c1998f69f86ae52a0dec23b977186e81/c/zss.c#L1192

That is, curves are specified as a string of 4 digit numbers back to back without any spaces or symbols inbetween.
Its very unfriendly to a human, so a convention of mapping from names to numbers is needed.
However, all of zowe needs to conform to the names. I dont know which name strings are available.

Node claims to use ASN.1 strings like "prime256v1"

So,
1. make a list of names and associated numbers
2. map from names to numbers to construct the number string (see https://www.ibm.com/docs/en/zos/3.1.0?topic=programming-cipher-suite-definitions#csdcwh__tttcsd )
3. in [tls.c](https://github.com/zowe/zowe-common-c/blob/v2.x/staging/c/tls.c), https://www.ibm.com/docs/en/zos/3.1.0?topic=reference-gsk-attribute-set-buffer must be used to set attribute "GSK_CLIENT_ECURVE_LIST" to the number string. if no number string is present, the command should be skipped as GSK defaults somehow.

JoeNemo:
This sounds like a straightforward parameter addition to the schema, and feeding that info through zss.c and zowe-common-c/c/tls.c.   

However the enumeration of valid values needs to be researched starting from [https://datatracker.ietf.org/doc/rfc4492/](url) and it's related and superseding RFC's.

1000TurquoisePogs:
I believe the full list of curve strings is here https://neuromancer.sk/std/x962/
Node wants the strings.
GSK wants the IANA numbers.
The IANA numbers are here https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml
Search within for "EC Named Curve Registry"

Short story: much like our TLS cipher implementation, we should utilize the name strings, and then create the string-to-number map.


You can even query what node supports, to see their syntax:
  'Oakley-EC2N-3',
  'Oakley-EC2N-4',
  'SM2',
  'brainpoolP160r1',
  'brainpoolP160t1',
  'brainpoolP192r1',
  'brainpoolP192t1',
  'brainpoolP224r1',
  'brainpoolP224t1',
  'brainpoolP256r1',
  'brainpoolP256t1',
[...]

PR #721, opened 2024-07-30
Curve customization support: extract curve array from yaml and make a list of mapped number strings

Gautham-coder:
Proposed changes
'zowe.network.server.tls.curves' is an array a user can set in zowe.yaml to customize crypto curves.
GSK handles curves as a string of 4 digit numbers(IANA numbers) back to back without any spaces or symbols in between.
Its very unfriendly to a human, so a mapping from names to numbers is needed, this is done in tls.h, for now only supported curves are in the array. Unsupported curves are commented and can be moved into this array as and when the supported curves are updated.

Curves are set during TLS settings initialization using gsk_attribute_set_buffer(), using 'GSK_CLIENT_ECURVE_LIST', see here
https://www.ibm.com/docs/en/zos/3.1.0?topic=reference-gsk-attribute-set-buffer for reference,

Currently, the supported curves are here, https://www.ibm.com/docs/en/zos/3.1.0?topic=programming-cipher-suite-definitions#csdcwh__tttcsd

Testing:
The curves that are not supported show an error in gsktrace as below, this was tested by adding some unsupported curves into the curve map array.
ERROR set_binary_ecurves(): Elliptical curve 0001 not supported
ERROR set_binary_ecurves(): Elliptical curve 0009 not supported
ERROR set_binary_ecurves(): Elliptical curve 0026 not supported

To show the error in normal logs, only valid curves are in the mapping array. So if any invalid curve is mentioned in zowe.yaml an invalid curve message is logged.

zowe.network.server.tls.curves: ["x25519", "x448", "secp192r1", "secp224r1","prime256v1","secp384r1", "secp521r1"]
,
is converted to number string 0029003000190021002300240025

This PR addresses Issue: https://github.com/zowe/zss/issues/713

This PR depends on: https://github.com/zowe/zowe-common-c/pull/466

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/c/zss.c
+++ b/c/zss.c
@@ -1224,38 +1224,46 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
   }
 
   Json *tlsConfig = NULL;
-  int tlsGetStatus = cfgGetAnyC(configmgr,ZSS_CFGNAME,&tlsConfig, 4,"zowe","network","server","tls");
+  int tlsGetStatus = cfgGetAnyC(configmgr, ZSS_CFGNAME, &tlsConfig, 4, "zowe", "network", "server", "tls");
   if (tlsGetStatus) {
     zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "TLS is NOT configured for this ZSS\n");
   } else {
     JsonObject *tlsConfigObject = jsonAsObject(tlsConfig);
     Json *curveJson = jsonObjectGetPropertyValue(tlsConfigObject, "curves");
     char *curves = NULL;
-    if(jsonIsArray(curveJson)) {
+    if (curveJson && jsonIsArray(curveJson)) {
       JsonArray *curveArray = jsonObjectGetArray(tlsConfigObject, "curves");
-      int count = jsonArrayGetCount(curveArray);
-      int curveCharLength = 4;
-      curves = (char *)safeMalloc((sizeof(char) * curveCharLength * count)+1, "curve list");
-      for (int i = 0; i < count; i++) {
-        char *ianaName = jsonArrayGetString(curveArray, i);
-        zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG, "curve request=%s\n", ianaName);
-        CurveMap *curve = (CurveMap *)ianaCurveMap;
-        bool found = false;
-        while (curve->groupId != NULL) {
-          if (!strcmp(ianaName, curve->name)) {
-            strcat(curves, curve->groupId);
-            zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG, "Curve match=%s\n", curve->groupId);
-            found = true;
-            break;
+      if (curveArray) {
+        int count = jsonArrayGetCount(curveArray);
+        const int curveCharLength = 4;
+        curves = safeMalloc((sizeof(char) * curveCharLength * count)+1, "curve list");
+        if (curves == NULL) {
+          zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, "Failure to allocate memory for Curves\n");
+          return false;
+        }
+        for (int i = 0; i < count; i++) {
+          char *ianaName = jsonArrayGetString(curveArray, i);
+          if (ianaName) {
+            zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG, "curve request=%s\n", ianaName);
+            CurveMap *curve = (CurveMap *)ianaCurveMap;
+            bool found = false;
+            while (curve->groupId != NULL) {
+              if (!strcmp(ianaName, curve->name)) {
+                strcat(curves, curve->groupId);
+                zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG, "Curve match=%s\n", curve->groupId);
+                found = true;
+                break;
+              }
+              ++curve;
+            }
+            if (!found) {
+              zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, ZSS_LOG_CURVE_INVALID_MSG, ianaName);
+            }
           }
-          ++curve;
         }
-        if (!found) {
-          zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_WARNING, ZSS_LOG_CURVE_INVALID_MSG, ianaName);
-        }
+        zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG, "Curve array is %s\n", curves);
+        settings->curves = curves;
       }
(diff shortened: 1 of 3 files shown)


c/zss.c
```diff
@@ -1226,6 +1227,42 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
     
   }
 
+  Json *tlsConfig = NULL;
+  int tlsGetStatus = cfgGetAnyC(configmgr,ZSS_CFGNAME,&tlsConfig, 4,"zowe","network","server","tls");
+  if (tlsGetStatus) {
+    zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "TLS is NOT configured for this ZSS\n");
```
ifakhrutdinov, line 1229: Should this have a message ID?

c/zss.c
```diff
@@ -1226,6 +1227,42 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
...
+  } else {
+    JsonObject *tlsConfigObject = jsonAsObject(tlsConfig);
+    Json *curveJson = jsonObjectGetPropertyValue(tlsConfigObject, "curves");
+    char *curves = NULL;
+    if(jsonIsArray(curveJson)) {
```
ifakhrutdinov: What if there is no `curves` in that JSON? I think you'll dereference a `NULL` pointer.

c/zss.c
```diff
@@ -1226,6 +1227,42 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
...
+      JsonArray *curveArray = jsonObjectGetArray(tlsConfigObject, "curves");
```
ifakhrutdinov, line 1235: Again, this can be `NULL`.

c/zss.c
```diff
@@ -1226,6 +1227,42 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
     
   }
 
+  Json *tlsConfig = NULL;
+  int tlsGetStatus = cfgGetAnyC(configmgr,ZSS_CFGNAME,&tlsConfig, 4,"zowe","network","server","tls");
```
ifakhrutdinov: Why do you in some cases use blanks between arguments and in some cases you don't?

c/zss.c
```diff
@@ -1226,6 +1227,42 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
...
+    zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "TLS is NOT configured for this ZSS\n");
+  } else {
+    JsonObject *tlsConfigObject = jsonAsObject(tlsConfig);
+    Json *curveJson = jsonObjectGetPropertyValue(tlsConfigObject, "curves");
+    char *curves = NULL;
+    if(jsonIsArray(curveJson)) {
+      JsonArray *curveArray = jsonObjectGetArray(tlsConfigObject, "curves");
+      int count = jsonArrayGetCount(curveArray);
+      int curveCharLength = 4;
+      curves = (char *)safeMalloc((sizeof(char) * curveCharLength * count)+1, "curve list");
```
ifakhrutdinov: `NULL` check missing?

ifakhrutdinov: Why are you casting this to `char *`?

c/zss.c
```diff
@@ -1226,6 +1227,42 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
...
+      for (int i = 0; i < count; i++) {
+        char *ianaName = jsonArrayGetString(curveArray, i);
```
ifakhrutdinov: Can this be `NULL`?

c/zss.c
```diff
@@ -1226,6 +1227,42 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
...
+  if (tlsGetStatus) {
+    zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_INFO, "TLS is NOT configured for this ZSS\n");
+  } else {
+    JsonObject *tlsConfigObject = jsonAsObject(tlsConfig);
+    Json *curveJson = jsonObjectGetPropertyValue(tlsConfigObject, "curves");
+    char *curves = NULL;
+    if(jsonIsArray(curveJson)) {
+      JsonArray *curveArray = jsonObjectGetArray(tlsConfigObject, "curves");
+      int count = jsonArrayGetCount(curveArray);
+      int curveCharLength = 4;
```
ifakhrutdinov: Make this `const` to indicate intention.

ifakhrutdinov: Why is it 4? Can you tie it to some struct's field size?

c/zss.c
```diff
@@ -1226,6 +1227,42 @@ static bool readAgentHttpsSettingsV2(ShortLivedHeap *slh,
...
+      curves = (char *)safeMalloc((sizeof(char) * curveCharLength * count)+1, "curve list");
+      for (int i = 0; i < count; i++) {
+        char *ianaName = jsonArrayGetString(curveArray, i);
+        zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG, "curve request=%s\n", ianaName);
+        CurveMap *curve = (CurveMap *)ianaCurveMap;
+        bool found = false;
+        while (curve->groupId != NULL) {
+          if (!strcmp(ianaName, curve->name)) {
+            strcat(curves, curve->groupId);
```
ifakhrutdinov: How do you know `groupId` is exactly 4?

ifakhrutdinov (changes requested):

c/zss.c
Gautham-coder: We know that the length will always be 4. Its better not to put in the struct field because we are initializing the array of structs in `zowe-common-c/h/tls.h` , but we can make it as a constant as you recommended earlier.

Gautham-coder: sure Irek.

c/zss.c
Gautham-coder: Yes that's true, thanks. Ill correct it.

c/zss.c
Gautham-coder: Yes Irek that's true, thanks for pointing that one.

c/zss.c
Gautham-coder: I'm not sure on that because elsewhere is the code I can see its a similar message without any ID as such.
Thanks

c/zss.c
Gautham-coder: Thanks for pointing that out Irek, I will make correct this.

c/zss.c
Gautham-coder: Thanks Irek, I will correct it.

Gautham-coder: Oh I think this is redundant, I will remove the typecast. Thanks

c/zss.c
Gautham-coder: Here: (https://www.ibm.com/docs/en/zos/3.1.0?topic=reference-gsk-attribute-set-buffer)
image
 
It says 1 or more 4-character decimal values.

c/zss.c
Gautham-coder: It is defined as a string so it can be NULL,  I will add a check for this. 
Thanks

Gautham-coder:
PR will be rebased with v3 since its an enhancement.

PR #726, opened 2024-08-20
Do not log into `zowe.runtimeDirectory`

Martin-Zeithaml:
## Proposed changes

This PR addresses Issue: https://github.com/zowe/zowe-install-packaging/issues/3495

## Type of change
- [x] Refactor the code 
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## PR Checklist

### `zowe.logDirectory` not used in config:
```
No log directory. Logging disabled.     
ZWED_NODE_LOG_FILE=/dev/null            
...
No log directory. Logging disabled.    
ZWES_LOG_FILE=/dev/null                
```

### `zowe.logDirectory` set to `/zowe/logs/`
```
ZWED_NODE_LOG_FILE=/zowe/logs/appServer-2024-08-27-07-39.log  
ZWES_LOG_FILE=/zowe/logs/zssServer-2024-08-27-07-39.log   
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to the ZSS package will be documented in this file.
 
 ## `3.0.0`
-
+- Bugfix: ommiting `zowe.logDirectory` forces the zss to log into the `zowe.runtimeDirectory`. (#726)
 
 ## `2.18.0`
 - Change log level for setting default value of 'httpRequestHeapMaxBlocks' to DEBUG instead of INFO.(#719)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to the ZSS package will be documented in this file.
 
 ## `3.0.0`
-- Enhancement: if no `zowe.logDirectory` is provided, logging is disabled. (#726)
+- Enhancement: if no `zowe.logDirectory` is defined in config, logging is disabled. (#726)
 
 ## `2.18.0`
 - Change log level for setting default value of 'httpRequestHeapMaxBlocks' to DEBUG instead of INFO.(#719)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to the ZSS package will be documented in this file.
 
 ## `3.0.0`
-- Bugfix: ommiting `zowe.logDirectory` forces the zss to log into the `zowe.runtimeDirectory`. (#726)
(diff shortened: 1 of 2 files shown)


1000TurquoisePogs:
You say "ommiting `zowe.logDirectory` forces the zss to log into the `zowe.runtimeDirectory`"
But that was by design.
It's intended for development: if you don't have a log directory, where should we log? We decided to log in the same location.

So yes you can change this behavior, but it sounds like the bug to me is actually: why does a user not have zowe.logDirectory ?
Sounds like something the schema should prevent?

Martin-Zeithaml:
1) The `zowe.logDirectory` is not mandatory and it is not in `defaults.yaml`. If you simply delete/comment out `logDirectory` line from config, it is valid from the schema point of view and `app-server` and `zss` are logging into `zowe.runtimeDirectory`.
2) I have created [this issue](https://github.com/zowe/zen/issues/243) to warn users from using `zowe.runtimeDirectory`.
a) If this check will be implemented, we might check (just for sure) all directories. It is possible, some new users locate certificates into `zowe.runtimeDirectory`.
3) Making `zowe.logDirectory` mandatory in schema validation is too aggressive - you must have `logDirectory` for every `zwe` command.
4) I don't know, if schema is capable of such complicated check, what we really need is to compare `"os.realpath(*Directory) vs os.realpath(runtimeDirectory)"` - you can define there symbolic links, relative/absolute path and who knows what else. 
5) At this point, you can define `zowe.logDirectory` and later comment it out - this check is better to do during the startup

## Possible solution in launcher?
We make it the same as for `zowe.workspaceDirectory`:
* `zowe.workspaceDirectory` is not mandatory, but you cannot run zowe without it:
```
ZWESVUSR INFO ZWEL0069I Configuration is valid             
ZWESVUSR ERROR ZWEL0059E failed to get WORKSPACE_DIR dir   
```
* We could possibly set `/dev/null` as default
  * By default, the logging is off, if you want to log, specify the directory.
  * **Edit:** Internal start prepare is trying to create `/dev/null` and fails, this should be already checked, since setting `zowe.logDirectory=/dev/null` is allowed
* Launcher could do the `realpath` check and either issue warning or stop if the `extensionDirectory | workspaceDirectory | logDirectory` is nested in `runtimeDirectory`
* Following setting will be illegal/not recommended:
```
  logDirectory: ${{ zowe.runtimeDirectory }}/logs
  workspaceDirectory: ${{ zowe.runtimeDirectory }}/workspace
  extensionDirectory: ${{ zowe.runtimeDirectory }}/extension
```

JoeNemo (approved):
Approved.

PR #750, opened 2025-03-03
Temporary fix for AT-TLS Gateway SSO negotiation

1000TurquoisePogs:
For SSO to work with ZSS, It needs to contact the gateway URL /gateway/api/v1/auth/keys/public/current
Currently, it does this only when ZSS is set up as an HTTPS server, because it sends an HTTPS request to the gateway and reuses the TLS settings.

This means when AT-TLS is enabled, currently, ZSS does not contact gateway at all. No SSO.

In this PR, I change that by
* Removing the requirement that the server must be setup for HTTPS in order to contact the gateway
* Making the gateway request conditionally HTTP / HTTPS depending upon the AT-TLS configuration

Switching to HTTP requests causes a little problem which I'm temporarily working around - the HTTP client code has different socket library behavior than the HTTPS client code (GSK).
In our HTTP client code, it seems that it's not blocking when it should, leading to us reading before the response is ready.
To work around this temporarily, I detect the error (EWOULDBLOCK) and do a sleep for 1 second before retrying, up to 9 seconds.
The SSO logic retries every 10 seconds until successful so, basically it may have an action every 1 second.

I will make another PR if I find a proper fix to the blocking behavior to avoid the 1 second polling.

### DIFF
--- a/c/jwk.c
+++ b/c/jwk.c
@@ -38,6 +38,7 @@
 #include "zssLogging.h"
 #include "jwt.h"
 #include "jwk.h"
+#include <unistd.h>
 
 static Json *receiveResponse(ShortLivedHeap *slh, HttpClientContext *httpClientContext, HttpClientSession *session, int *statusOut);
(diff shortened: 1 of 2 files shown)


JoeNemo (approved):
Approve

PR #757, opened 2025-03-24
Update proper trace name for JWT tracing

1000TurquoisePogs:
I needed to trace the JWT code of zowe-common-c and found the schema did not match the behavior
ZSS calls the set function for JWT tracing when it sees the logLevel "_zss.jwtTrace", not "_zss.jwt": https://github.com/zowe/zss/blob/06c9143b3421ed3b9d5c55d8e63979e75f3160a6/c/zss.c#L452


JoeNemo (approved):
Approved.

PR #763, opened 2025-04-10
Update incorrect CHANGELOG.md number

1000TurquoisePogs:
v2.18.1 changelog has an item which points to a ticket number that is unrelated.
I have identified the PR that this changelog message came from, and have set it to reference that PR number instead.


ifakhrutdinov (approved):
LGTM

PR #764, opened 2025-04-11
Create "zis-test" and "bind-test" simple utilities for pre-flight checks

1000TurquoisePogs:
This PR adds two tools that can be used by users or by zowe to validate that some Zowe requirements are met before doing an action such as starting Zowe.

`zis-test --zis <crossMemoryServerName>`: this utility tries to connect to ZIS, and using the cmsRC, suggests actions the user should take if it is not successful, and returns an RC of either 0 or 8 to indicate success/fail.

`_BPX_JOBNAME=<jobname> bind-test --host <hostname> --port <port>`: this utility tries to start a simple tcp server on a given host/port combination using a given jobname, to detect if TCPIP rules allow/restrict this combination. This also can detect if a port is already occupied by something else.

"zis-test" is now part of zss validate.sh, so that startup will be prevented if it detects that ZIS cant be contacted.


c/zisTest.c
```diff
@@ -0,0 +1,100 @@
...
+
+int main(int argc, char **argv){
+  int status = VERIFY_STATUS_OK;
+
+  char *zisName = getKeywordArg("--zis",argc,argv);
+
+  if (!zisName) {
+    printf("Error: --zis specifying ZIS server name to check is required\n");
+    status = VERIFY_STATUS_ERROR;
+  } else if (strlen(zisName) > 16) {
```
ifakhrutdinov: Please don't use hardcoded lengths as it's error-prone.
You can do:
```c
if (strlen(zisName) > sizeof(CrossMemoryServerName)) {
  printf("Error: ZIS server name must be maximum %zu characters.\n",
         sizeof(CrossMemoryServerName));
```

c/zisTest.c
```diff
@@ -0,0 +1,100 @@
...
+  Copyright Contributors to the Zowe Project.
+*/
+
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <stdarg.h>
+#include <sys/stat.h>
+#include <iconv.h>
```
ifakhrutdinov: Why do you need all these headers?

I think you can get by with this:
```c
#include <stdio.h>
#include <string.h>
#include <stdlib.h>

#include "zis/client.h"
```

c/zisTest.c
```diff
@@ -0,0 +1,100 @@
...
+      return argv[aa+1];
+    }
+  }
+  return NULL;
+}
+
+#define VERIFY_STATUS_OK     0
+#define VERIFY_STATUS_ERROR  8
+
+int main(int argc, char **argv){
```
ifakhrutdinov: Though not critical for this PR, I'd work on the consistency of the style because it's all over. I would use an IDE or a plug-in for your editor which can autoformat.

build/build_zis_test.sh
```diff
@@ -0,0 +1,65 @@
...
+
+echo "********************************************************************************"
+echo "Building zis check ..."
+
+
+mkdir -p "${WORKING_DIR}/tmp-zis-test" && cd "$_"
+
+export _C89_ACCEPTABLE_RC=0
+
+if ! c89 \
```
ifakhrutdinov: You can simplify this:
```diff
diff --git i/build/build_zis_test.sh w/build/build_zis_test.sh
index ba211ca..ed5d947 100755
--- i/build/build_zis_test.sh
+++ w/build/build_zis_test.sh
@@ -12,12 +12,7 @@ set -e
 ################################################################################
 
 
-export _C89_LSYSLIB="CEE.SCEELKED:SYS1.CSSLIB:CSF.SCSFMOD0"
-export _C89_L6SYSLIB="CEE.SCEEBND2:SYS1.CSSLIB:CSF.SCSFMOD0"
-
 WORKING_DIR=$(cd $(dirname "$0") && pwd)
-ZSS_ROOT="$WORKING_DIR/.."
-COMMON_BUILD="$WORKING_DIR/../deps/zowe-common-c/build"
 ZSS="../.."
 COMMON="../../deps/zowe-common-c"
 
@@ -29,27 +24,19 @@ echo "Building zis check ..."
 
 mkdir -p "${WORKING_DIR}/tmp-zis-test" && cd "$_"
 
-export _C89_ACCEPTABLE_RC=0
-
 if ! c89 \
   -D_XOPEN_SOURCE=600 \
-  -D_OPEN_THREADS=1 \
-  -DAPF_AUTHORIZED=0 \
+  -DCMS_CLIENT \
   -DNEW_CAA_LOCATIONS=1 \
-  -Wc,lp64,expo,langlvl\(extc99\),gonum,goff,hgpr,roconst,ASM,asmlib\('CEE.SCEEMAC','SYS1.MACLIB','SYS1.MODGEN'\) \
+  -Wc,lp64,langlvl\(extc99\),gonum,goff,hgpr,roconst,ASM,asmlib\('SYS1.MACLIB'\) \
   -Wc,agg,exp,list,so\(\),off,xref \
   -Wl,lp64 \
   -I ${COMMON}/h \
   -I ${ZSS}/h \
   -o ${ZSS}/bin/zis-test \
   ${COMMON}/c/alloc.c \
-  ${COMMON}/c/collections.c \
   ${COMMON}/c/crossmemory.c \
-  ${COMMON}/c/le.c \
-  ${COMMON}/c/logging.c \
   ${COMMON}/c/zos.c \
-  ${COMMON}/c/recovery.c \
-  ${COMMON}/c/scheduling.c \
   ${COMMON}/c/timeutls.c \
   ${COMMON}/c/utils.c \
   ${COMMON}/c/zvt.c \
```

ifakhrutdinov: I think you need to drop `!` because, afaik is sh, if the command is successful in `if` (i.e., return code 0), the `then` branch is executed.

ifakhrutdinov (commented):
@1000TurquoisePogs, please see my comments; they're also applicable to the c/sh files where I didn't comment.

c/zisTest.c
1000TurquoisePogs: What style is bothering you?
is it `){` vs `) {`?
Its all over the codebase from Joe & I, because we both use emacs on Z in a pinch and on Z that's emacs 19 from like 1990 or something and it does autoformat - to that.

ifakhrutdinov: I know and that's not a good thing, but you've created a new source file and you can at least keep it consistent (e.g., your example and inconsistent spacing between the function arguments). Ideally, we should add something like `.clang-format` and ask people apply it, or automate that.

As for what's bothering me, it doesn't matter since we don't really have style guidelines and this is a new file.

c/bindTest.c
```diff
@@ -0,0 +1,98 @@
...
+#define BIND_STATUS_OK     0
+#define BIND_STATUS_ERROR  8
+
+int main(int argc, char **argv){
+  int status = BIND_STATUS_OK;
+
+  int returnCode = 0;
+  int reasonCode = 0;
+  int port = atoi(getKeywordArg("--port",argc,argv));
+  char *address = getKeywordArg("--host",argc,argv);
```
Martin-Zeithaml, line 67: `--help` with some basic info would be nice, same for `zisTest.c`.

build/build_zis_test.sh
1000TurquoisePogs: that does not compile. i have not been able to remove any of the c files, they are all referenced by the zis code in some way.

1000TurquoisePogs: worked now, dont know why i had an issue before. fixed

1000TurquoisePogs:
I believe all issues are fixed now and ready to go.

c/bindTest.c
```diff
@@ -0,0 +1,103 @@
...
+    return TRUE;      
+  }
+  if (!(*inetAddress && (*inetAddress)->data.data4.addrBytes)) {
+    printf("Error: Could not resolve hostname from DNS. Is this a valid hostname for this system?\n");
+    return FALSE;
+  } else {
+    int part4 = 0xff & (*inetAddress)->data.data4.addrBytes;
+    int part3 = (0xff00 & (*inetAddress)->data.data4.addrBytes) >> 8;
+    int part2 = (0xff0000 & (*inetAddress)->data.data4.addrBytes) >> 16;
+    int part1 = (0xff000000 & (*inetAddress)->data.data4.addrBytes) >> 24;
```
1000TurquoisePogs, line 33: IPv6 support is coming soon, so I've filed https://github.com/zowe/zss/issues/770 as a reminder to enhance this at that time.

c/zisTest.c
```diff
@@ -0,0 +1,87 @@
...
+    return VERIFY_STATUS_OK;
+  }
+
+  char *zisName = getKeywordArg("--zis",argc,argv);
+
+  if (!zisName) {
+    printf("Error: --zis specifying ZIS server name to check is required\n");
+    status = VERIFY_STATUS_ERROR;
+  } else if (strlen(zisName) > sizeof(CrossMemoryServerName)) {
+    printf("Error: ZIS server name must be maximum 16 characters.\n");
```
ifakhrutdinov: Since you remove the hardcoded length from the `if` condition, why not remove it from `printf`?
```c
printf("Error: ZIS server name must be maximum %zu characters.\n", sizeof(CrossMemoryServerName));
```

c/zisTest.c
```diff
@@ -0,0 +1,87 @@
...
+    status = VERIFY_STATUS_ERROR;
+  } else {
+    int rc = printZISStatus(zisName);
+    if (rc != 0) {
+      status = VERIFY_STATUS_ERROR;
+      if (rc == RC_CMS_PERMISSION_DENIED) {
+        printf("Error: program lacks permission to use ZIS\n");
+        char *username = getenv("USER");
```
ifakhrutdinov, line 74: I think you can reliably get the user name which got tested by ZIS:
```c
#include "zos.h"

typedef struct UserName_tag {
  char valueNullTerm[9];
} UserName;

static UserName getAddressSpaceUserName(void) {
  UserName name = {0};
  const ASXB *asxb = getASXB();
  const ACEE *acee = (ACEE *)asxb->asxbsenv;
  unsigned userNameLen = acee->aceeuser[0];
  const char *userName = &acee->aceeuser[1];
  memcpy(name.valueNullTerm, userName,
         min(userNameLen, sizeof(name.valueNullTerm) - 1));
  return name;
}

UserName asUser = getAddressSpaceUserName();
printf("Address space user %s has no access to ZIS\n", asUser.valueNullTerm);
```

_I have not compiled this code_.

c/zisTest.c
```diff
@@ -0,0 +1,87 @@
...
+        if (username) {
+          printf("Ensure the Zowe STC id (possibly the current user: %s) has READ access to ZWES.IS in the FACILITY class\n", username);
+        } else {
+          printf("Ensure the Zowe STC id has READ access to ZWES.IS in the FACILITY class\n");
+        } 
+      } else if (rc == RC_CMS_ZVT_NULL || rc == RC_CMS_ZERO_PC_NUMBER || rc == RC_CMS_GLOBAL_AREA_NULL) {
```
ifakhrutdinov: I would add `RC_CMS_SERVER_NOT_READY` to this pile.

c/bindTest.c
```diff
@@ -0,0 +1,103 @@
...
+  }
+  
+  int tlsFlags = 0;
+  Socket *serverSocket = tcpServer2(inetAddress, port, tlsFlags, &returnCode, &reasonCode);
+   
+  if (serverSocket) {
+    printf("Bind succeeded (pointer=0x%p, rc=0x%x, rsn=0x%x)\n", serverSocket, returnCode, reasonCode);
+  } else{
+    status = BIND_STATUS_ERROR;
+    char *jobname = getenv("_BPX_JOBNAME");
```
ifakhrutdinov, line 87: See my other comments about the user name. Maybe you want to consider a more reliable way getting that info (and the job name) here too.

c/bindTest.c
```diff
@@ -0,0 +1,103 @@
...
+    printf("Error: Invalid address given for --host\n");
+    return BIND_STATUS_ERROR;
+  }
+  if (!port) {
+    printf("Error: No --port given\n");
+    return BIND_STATUS_ERROR;
+  }
+  
+  int tlsFlags = 0;
+  Socket *serverSocket = tcpServer2(inetAddress, port, tlsFlags, &returnCode, &reasonCode);
```
ifakhrutdinov, line 81: Should we close this socket when done?

ifakhrutdinov (commented):
Overall, it's ok, but see my comments in case you want to improve the code.

c/bindTest.c
1000TurquoisePogs: Does ZSS have such code?
I haven't observed this nor ZSS occupying sockets after terminating.
I will add it, but it's another thing to test if it has any impact.

c/zisTest.c
1000TurquoisePogs: USER env var is used elsewhere in Zowe for this purpose, but that may also work.

JoeNemo (approved):
Approve.

PR #772, opened 2025-06-11
Fix potential loop on HTTP/ATTLS Client JWK behavior

1000TurquoisePogs:
When HTTP/ATTLS is used in ZSS, when it tries to contact Gateway for SSO, the non-blocking loop for reading the gateway response was flawed. There's a wait loop that looks for conditions to wait, but the wait is way too small and there are conditions where the loop is infinite as well, causing a rapid and high-cpu loop:

after

This PR addresses it by making the wait between loops longer, unconditional, and handles the potential status codes better.

Here's the change - you'll see no CPU issue.
before

### DIFF
--- a/c/jwk.c
+++ b/c/jwk.c
@@ -211,15 +211,16 @@ static Json *receiveResponse(ShortLivedHeap *slh, HttpClientContext *httpClientC
   int currentLoop = 0;
   while (!done) {
     int status = httpClientSessionReceiveNativeLoop(httpClientContext, session);
-    if (status == HTTP_CLIENT_EWOULDBLOCK){
-      currentLoop++;
-      usleep(1000);
-    }
     if (currentLoop > loopLimit){
       zowelog(NULL, LOG_COMP_ID_JWK, ZOWE_LOG_WARNING, "JWT timeout reached\n");
       break;
     }
-    if (status != 0 && status != 15) {
+    if (status == HTTP_CLIENT_SOCKET_TIMEOUT) {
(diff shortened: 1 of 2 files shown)


c/jwk.c
```diff
@@ -211,22 +211,25 @@ static Json *receiveResponse(ShortLivedHeap *slh, HttpClientContext *httpClientC
...
     if (currentLoop > loopLimit){
       zowelog(NULL, LOG_COMP_ID_JWK, ZOWE_LOG_WARNING, "JWT timeout reached\n");
       break;
     }
-    if (status != 0 && status != 15) {
+    if (status == HTTP_CLIENT_SOCKET_TIMEOUT) {
+      zowelog(NULL, LOG_COMP_ID_JWK, ZOWE_LOG_WARNING, "socket timeout reached\n");
+      break;
+    } else if (status != 0 &&
+               status != HTTP_CLIENT_READ_ERROR &&
```
ifakhrutdinov: Why don't you want to break out of the loop if there is a read error?

ifakhrutdinov:
@JoeNemo, this PR had the same issue as the [v2 PR](https://github.com/zowe/zss/pull/775), and I've picked up t[he same fix](https://github.com/zowe/zss/pull/772/commits/d41992a9afabfd9fd2e5b6534393ff4cd33fd624) I did for the v2 PR. Since this has now my commit, I think you should review it.

JoeNemo (approved):
Approved.

PR #787, opened 2025-07-30
Accomodate components.apiml.enabled=true

1000TurquoisePogs:
This PR checks all variables that used to check for discovery/gateway/caching-service and allows components.apiml.enabled to be a substitute for them.

Similar to https://github.com/zowe/zlux-app-server/pull/345

### DIFF
--- a/defaults.yaml
+++ b/defaults.yaml
@@ -28,13 +28,13 @@ components:
       mediationLayer:
         server:
           isHttps: true
-          enabled: ${{ components.gateway.enabled }}
+          enabled: ${{ components.gateway?.enabled || components.apiml?.enabled }}
           gatewayHostname: ${{ std.getenv('ZWE_haInstance_hostname') }}
           hostname: ${{ std.getenv('ZWE_haInstance_hostname') }}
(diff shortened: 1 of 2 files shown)


defaults.yaml
```diff
@@ -28,13 +28,13 @@ components:
       mediationLayer:
         server:
           isHttps: true
-          enabled: ${{ components.gateway.enabled }}
+          enabled: ${{ components.gateway?.enabled || components.apiml?.enabled }}
           gatewayHostname: ${{ std.getenv('ZWE_haInstance_hostname') }}
           hostname: ${{ std.getenv('ZWE_haInstance_hostname') }}
           gatewayPort: ${{ components.gateway.port }}
           port: ${{ components.discovery.port }}
```
pablocarle, line 35: It's not documented and it's still open for discussion but for the future (possible v4), these ports and settings in general can be defined under the `apiml` component

MarkAckert (approved):

defaults.yaml
1000TurquoisePogs: ok - just let me know when that is certain so more adjustments can be made.

PR #788, opened 2025-07-31
Add name to zis test output

1000TurquoisePogs:
This PR just adds the zis name to the output of the zis test, so users are aware of what to look for.


Gautham-coder (approved):
Changes look good.

PR #807, opened 2025-11-19
Improve JWK messages

1000TurquoisePogs:
This is a PR that takes a pragmatic approach about that we know what follow-up troubleshooting certain JWK error messages involve, so we should just print more about what happened and what the user can do about it.


JoeNemo (approved):
Approved.

c/jwk.c
```diff
@@ -280,6 +288,12 @@ static void getPublicKey(Json *jwk, x509_public_key_info *publicKeyOut, int *sta
   if (!keyObject) {
     zowelog(NULL, LOG_COMP_ID_JWK, ZOWE_LOG_WARNING, "JWK doesn't contain key\n");
     *statusOut = JWK_STATUS_UNRECOGNIZED_FMT_ERROR;
+
+    zowelog(NULL, LOG_COMP_ID_JWK, ZOWE_LOG_WARNING, "JWK response:\n");
+    //Often enough, the destination has some error message that can be printed
+    jsonPrinter *jp = makeJsonPrinter(STDOUT_FILENO);
```
ifakhrutdinov, line 294: @1000TurquoisePogs , are we sure there is no sensitive information at this point?

h/zssLogging.h
```diff
@@ -518,8 +518,10 @@ bool isLogLevelValid(int level);
 #ifndef ZSS_LOG_JWK_FAILED_MSG_ID
 #define ZSS_LOG_JWK_FAILED_MSG_ID     ZSS_LOG_MSG_PRFX"1605W"
 #endif
-#define ZSS_LOG_JWK_FAILED_MSG_TEXT   "Server will not accept JWT\n"
+#define ZSS_LOG_JWK_FAILED_MSG_TEXT   "Server will not accept JWT\nCheck URL https://%s:%d%s for errors.\n"
```
ifakhrutdinov, line 521: I suggest we have a single messages: `Server will not accept JWT\nCheck URL %s for errors.\n` and format the path in a separate function. The reason for that is for a simple task (print an error message) we're spilling so much unnecessary details here and in `jwkTaskMain`; it's better to isolate that.

c/jwk.c
```diff
@@ -118,14 +118,22 @@ static int jwkTaskMain(RLETask *task) {
...
+        if (rsn == HTTP_CLIENT_TLS_ERROR) {
+          zowelog(NULL, LOG_COMP_ID_JWK, ZOWE_LOG_WARNING, "If TLS error persists, trace GSK for more detail using YAML property 'components.zss.agent.https.trace: true'\n");
+        }
       }
       sleep(retryIntervalSeconds);
     }
   }
   if (success) {
     zowelog(NULL, LOG_COMP_ID_JWK, ZOWE_LOG_INFO, ZSS_LOG_JWK_READY_MSG, settings->fallback ? "with" : "without");
   } else {
```
ifakhrutdinov, line 130: I'd add a helper function and simplify this code:
```c
static char *printJWKURL(const JwkSettings *settings, char *buffer, size_t bufferLen) {
  int charsPrinted;
  if (indexOf(settings->host, strlen(settings->host), ':', 0) != -1) {
    charsPrinted = snprintf(buffer, bufferLen, "https://[%s]:%d%s", settings->host, settings->port, settings->path);
  } else {
    charsPrinted = snprintf(buffer, bufferLen, "https://%s:%d%s", settings->host, settings->port, settings->path);
  }
  if (charsPrinted < 0) {
    memset(buffer, 0, bufferLen);
  }
  return buffer;
}

static int jwkTaskMain(RLETask *task) {
...
  if (success) {
    zowelog(NULL, LOG_COMP_ID_JWK, ZOWE_LOG_INFO, ZSS_LOG_JWK_READY_MSG, settings->fallback ? "with" : "without");
  } else {
    char url[256];
    zowelog(NULL, LOG_COMP_ID_JWK, ZOWE_LOG_WARNING, ZSS_LOG_JWK_FAILED_MSG, printJWKURL(settings, url, sizeof(url)));
  }
...
}
```

c/jwk.c
```diff
@@ -280,6 +288,12 @@ static void getPublicKey(Json *jwk, x509_public_key_info *publicKeyOut, int *sta
   if (!keyObject) {
     zowelog(NULL, LOG_COMP_ID_JWK, ZOWE_LOG_WARNING, "JWK doesn't contain key\n");
     *statusOut = JWK_STATUS_UNRECOGNIZED_FMT_ERROR;
+
+    zowelog(NULL, LOG_COMP_ID_JWK, ZOWE_LOG_WARNING, "JWK response:\n");
```
Martin-Zeithaml, line 292: Probably needs `fflush(stdout);` after the `zowelog`. Otherwise I can see message header after the text:
```
"possible error message"JWK response:
```
instead of
```
JWK response:
"possible error message"
```

ifakhrutdinov: On top of that, the message might get split by some other stray message since they're not printed as a single message.

ifakhrutdinov (commented):
See my recommendation (I also incorporated them in a [commit](https://github.com/zowe/zss/commit/a3c83fdc43fca0a0e391c49aa2a2740f5198e663), which has only been build-tested).

c/jwk.c
1000TurquoisePogs: It's information you can get in the browser without authentication, so yeah I don't see how it can be sensitive. It's a public key used to check authenticity of later JWT

c/jwk.c
1000TurquoisePogs: This looks to make the code more complicated. Why do this?

ifakhrutdinov: > Why do this?

Because it's not more complicated?

In the current implementation, you mix lower and higher level logic, and then you have to introduce additional message templates just to work around the peculiarities of IPv6; this all basically unnecessarily spreads lower level details over a larger area than needed.

So, why not just hide this in a specialised function which can be easily reviewed and verified?

PR #811, opened 2026-03-09
APIML Eureka registration

1000TurquoisePogs:
This PR adds an http client that implements eureka registration to apiml.
The alternative and prior behavior was to use a static registration file. This is retained under the yaml option `components.zss.agent.mediationLayer.static: true`.

Timing parameters regarding the registration & heartbeat loop have been added to the YAML and schema.

This is a draft until I get to testing it.


pablocarle:
Hi, just a quick question, will this handle migration scenarios? like what happens if the workspace has a static definition file for zss and dynamic onboarding is enabled? will it delete the static definition file?

1000TurquoisePogs:
@pablocarle yes this is handled in configure.sh
zss handles cleanup of whether the static file should exist or not when it runs.

if zss is disabled, it wont run this logic so instead the zwe startup code must handle it here https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/bin/libs/component.ts#L638-L673 

I am making the reverse of this PR for app-server.
zss had static only, now it will have static or eureka
app-server has eureka, now it will have static or eureka.
the goal is to make HA better but also simplify networking for users who have hard to diagnose issues.

1000TurquoisePogs:
I have now tested that this code works for normal TLS, but haven't yet tested ATTLS.
I recall there's this bug in apiml that it needs to be faked out about whether to make http/https calls so I believe I'll need to add the logic seen here next https://github.com/zowe/zlux-server-framework/blob/v3.x/staging/lib/apiml.js#L198-L218

PR #825, opened 2026-04-23
Allow encoding preference on get calls

1000TurquoisePogs:
## Proposed changes

This PR addresses that USS file content endpoint should accept encoding name strings and integer CCSIDs as query parameters).

Depends upon https://github.com/zowe/zowe-common-c/pull/593

---

### Summary

**`getEncodingInfoFromQueryParameters()` in `c/unixFileService.c`**

The PUT path for `/unixfile/contents/**` reads `sourceEncoding` and `targetEncoding` query parameters to govern character conversion during upload. Previously these were resolved with `getCharsetCode()`, which only recognised a small set of named charset strings and returned -1 for integer strings such as `"1047"`.

The calls have been replaced with `parseEncodingValue()` (new in zowe-common-c), which accepts:
- Named charset strings: `"IBM-1047"`, `"UTF-8"`, `"IBM-819"`, and all IANA IBM codepage names in `"IBM-NNNN"` format
- Decimal CCSID integer strings: `"37"`, `"819"`, `"1047"`, `"1208"`, etc. (range 1–65535)
- The sentinel `"binary"` (case-insensitive) → `CCSID_BINARY` (65535)

The GET path encoding parameters (`sourceEncoding`/`targetEncoding`) are handled inside `respondWithUnixFile2()` in zowe-common-c's `httpserver.c` and required no changes in this repository.

## Type of change

## PR Checklist

- [x] My code follows the style guidelines of this project

## Testing

1. **PUT with named charset**: `PUT /unixfile/contents/path/file.c?sourceEncoding=IBM-1047&targetEncoding=UTF-8` should upload and convert correctly.
2. **PUT with integer CCSID**: `PUT /unixfile/contents/path/file.c?sourceEncoding=1047&targetEncoding=1208` should produce the same result as test 1.
3. **PUT with mixed**: `PUT /unixfile/contents/path/file.c?sourceEncoding=IBM-1047&targetEncoding=1208` should work.
4. **PUT with invalid value**: `PUT /unixfile/contents/path/file.c?sourceEncoding=bogus` should return HTTP 400.
5. **PUT without params**: existing default behaviour preserved.

## Further comments

This change is strictly additive — existing callers that already supply integer strings continue to work, and callers that now supply named strings gain new functionality without any API break.


JoeNemo:
I resolved the conflicts and pushed two commits, then went looking for what the
one-line swap actually changes. Summary: the change is right and worth having,
but it is not strictly additive -- it drops an input rejection -- and I think
that wants a decision before merge.

### What I pushed

- `08c0e7e` -- merge of `v3.x/staging`. Only two files conflicted: the
  `deps/zowe-common-c` pointer (taken from staging, `e66184d`, which is ahead of
  what this branch carried) and `CHANGELOG.md` (kept both sides). GitHub now
  reports the branch mergeable.
- `d23d6ba` -- the binary sentinel, explained below.

### Measured before/after

I lifted `getEncodingInfoFromQueryParameters` verbatim into a harness and linked
it against the real `charsets.c`/`utils.c` at `e66184d`, so only the changed
line differs between the two runs. Same inputs, both forms:

```
                                BEFORE (getCharsetCode)   AFTER (parseEncodingValue)
  source=IBM-1047 target=UTF-8   1047 / 1208  accepted     1047  / 1208  accepted
  source=1047     target=1208      -1 /   -1  400          1047  / 1208  accepted   <- the point of the PR
  source=IBM-1047 target=1208     1047 /   -1  400         1047  / 1208  accepted   <- ditto
  source=binary   target=binary     -1 /   -1  accepted      -1  /   -1  accepted
  source=binary   target=UTF-8      -1 / 1208  400         65535  / 1208  accepted   <- new
  source=UTF-8    target=binary   1208 /   -1  400          1208  / 65535 accepted   <- new
  source=bogus    target=UTF-8      -1 / 1208  400            -1  / 1208  400
```

The middle two rows are the PR's purpose and they work. The last two are the
part I would not have predicted from the description.

### Mixed binary/text is no longer rejected

`parseEncodingValue` accepts `binary` as an ordinary encoding name, so a request
naming `binary` on one side and a charset on the other no longer trips the `-1`
check in `parseEncodingParameter`. It is classified `TEXT` and admitted, and
then in `assignSessionIDToCaller` both values are used:

- `tagFileForUploading` (line 519) runs unconditionally, so `target=binary`
  tags the file `0xFFFF` -- binary -- while the session is in TEXT mode.
- `writeAsciiDataFromBase64` (line 636) is handed `65535` as a source or target
  CCSID and asked to convert to or from it.

Previously that combination was a clean 400. I do not think anyone wants the new
behaviour, but it is a contract change either way, so it should be deliberate.
The narrow fix is to reject it explicitly once the values are parsed, which
restores the old outcome without giving up the integer support:

```c
if (*outTransferType == TEXT &&
    ((unsigned short)*outSourceCCSID == (unsigned short)CCSID_BINARY ||
     (unsigned short)*outTargetCCSID == (unsigned short)CCSID_BINARY)) {
  /* binary is all-or-nothing; pairing it with a charset has no meaning */
  respondWithJsonError(response, "binary may only be paired with binary", 400, "Bad Request");
  return -1;
}
```

Happy to push that if you agree it is the behaviour you want.

### The binary sentinel (`d23d6ba`)

While tracing the above I hit a collision. `CCSID_BINARY` is `(short)0xFFFF`, so
storing it in an `int` sign-extends to `-1` -- the same value
`parseEncodingParameter` reads as "this encoding could not be parsed". Before
this PR that was invisible, because reaching the binary branch required both
parameters to be spelled `binary` and every mixed case was rejected anyway.
After it, the same sentinel exists in two representations at once: `binary` on
one side yields `65535`, `binary` on both yields `-1`.

I cast at the assignment the way `parseEncodingValue` already does, so it is
`65535` everywhere and `-1` again means only "parse failed" (bottom row of the
table above). `CCSID_BINARY` is never the right-hand side of a comparison
anywhere in either repo, so nothing else reads the widened value.

The same shape exists in `patternChangeTagTest` in zowe-common-c, where it is
still latent -- filed as zowe/zowe-common-c#665 rather than folded in here.

### Two small things

- **Title.** "Allow encoding preference on get calls" -- the only caller of
  `parseEncodingParameter` is `assignSessionIDToCaller`, which is the upload
  session path. No GET path is touched. The CHANGELOG entry says PUT, so I
  think the title is just stale.
- **CHANGELOG.** Two corrections. The direction is reversed: it reads
  "Previously only integer strings were accepted", but the old `getCharsetCode`
  accepted *names* and rejected integers -- rows 1 and 2 of the table. And the
  link resolves to zowe/zss#593, a merged PR about `zisReadParmlib` RCs. The
  intended reference is presumably zowe/zowe-common-c#593, which is closed and
  was superseded by zowe/zowe-common-c#630.

1000TurquoisePogs:
I approve of the fixes @JoeNemo  added.

JoeNemo (approved):
Approved.

PR #831, opened 2026-05-08
Make alloc test binary

1000TurquoisePogs:
## Proposed changes

Add the `alloc-test` binary: an empirical memory probe for z/OS that tests both the 64-bit (above-the-bar) and 31-bit (below-the-bar) address-space regions. This binary is used by the new `zwe validate memory` command (in `zowe-install-packaging`) to confirm that allocations of the expected sizes actually succeed at runtime, independent of ESM configuration values.

## Type of change

### New files

**`c/allocTest.c`**
- Probes available memory in two z/OS address-space regions:
  - 64-bit (above 2 GB bar) via `malloc()` — gated by `MEMLIMIT`
  - 31-bit (below 2 GB bar) via `malloc31()` from `zowe-common-c/c/alloc.c` — gated by `ASSIZEMAX` / `ulimit -A`
- Accepts `--max64`, `--max31`, and `--min` size arguments (e.g. `2G`, `256M`, `32M`)
- Enforces a 16 G absolute ceiling (`MAX_PROBE_BYTES`) on both max arguments
- Uses a decreasing power-of-2 loop; no `memset` is performed, so probes are near-instant even for multi-gigabyte sizes
- Outputs one line per region: `64-bit allocation succeeded at 2G` / `31-bit allocation succeeded at 256M` (or a failure message with a diagnostic hint)
- Exit code 0 if both regions succeed; 8 if either fails
- Links `zowe-common-c/c/alloc.c` and `zowe-common-c/h/alloc.h`

## PR Checklist

## Testing

```sh
# Both regions should succeed and print sizes
bin/alloc-test --max64 2G --max31 256M --min 32M

# Expected output (sizes may vary by MEMLIMIT / ASSIZEMAX configuration):
# 64-bit allocation succeeded at 2G
# 31-bit allocation succeeded at 256M

# Test argument validation
bin/alloc-test --max64 32G   # should fail: exceeds 16G ceiling
bin/alloc-test --max64 1G --min 2G  # should fail: --min >= --max64
```

## Further comments

- `malloc31()` (from `zowe-common-c/c/alloc.c`) calls `__malloc31()` directly without `memset`, making it safe for probing large sizes without touching pages. `safeMalloc31()` was deliberately avoided because it calls `memset(res, 0, size)` after allocation, which would cause a 256 M probe to take more time.
- Similarly, `malloc()` (not `safeMalloc64()`) is used for the 64-bit probe for the same reason.
- The 16 G absolute ceiling (`MAX_PROBE_BYTES`) prevents runaway probes regardless of what arguments are passed.

### DIFF
--- a/.pax/alloc-test/pre-packaging.sh
+++ b/.pax/alloc-test/pre-packaging.sh
@@ -1,0 +1,34 @@
+#!/bin/sh -e
+set -xe
+
+################################################################################
+# This program and the accompanying materials are made available under the terms of the
+# Eclipse Public License v2.0 which accompanies this distribution, and is available at
+# https://www.eclipse.org/legal/epl-v20.html
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Copyright Contributors to the Zowe Project.
+################################################################################
+
+
+# contants
+SCRIPT_NAME=$(basename "$0")
+SCRIPT_DIR=$(pwd)
+
+# build
+echo "$SCRIPT_NAME build alloc-test ..."
+STEPLIB=CBC.SCCNCMP "$SCRIPT_DIR/content/build/build_alloc_test.sh"
+if [ $? != 0 ]; then
+  exit $?
+fi
+
+# clean up content folder
+echo "$SCRIPT_NAME cleaning up pax folder ..."
+cd "$SCRIPT_DIR"
+mv content bak && mkdir -p content
+
+# move real files to the content folder
+echo "$SCRIPT_NAME coping files should be in pax ..."
+cd "$SCRIPT_DIR/content"
+cp ../bak/bin/alloc-test .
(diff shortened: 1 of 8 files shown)


.pax/alloc-test/pre-packaging.sh
```diff
@@ -0,0 +1,34 @@
...
+  exit $?
+fi
+
+# clean up content folder
+echo "$SCRIPT_NAME cleaning up pax folder ..."
+cd "$SCRIPT_DIR"
+mv content bak && mkdir -p content
+
+# move real files to the content folder
+echo "$SCRIPT_NAME coping files should be in pax ..."
```
Martin-Zeithaml, line 32: coping -> copying?

ifakhrutdinov:
I haven't yet looked at the code, but I suspect that this test can be both false positive and false negative because of how the virtual storage limits are set (e.g., via an STC JCL). That's not to defeat the purpose of this program, but we should still keep that in mind.

> independent of ESM configuration values.

Can ESMs impose memory limits? If yes, how?

PR #833, opened 2026-05-28
Fix target length condition in `CMSModifyCommand`

Martin-Zeithaml:
* `memcpy(nickname.text, command->target, targetLength);`
* `nickname.text` is destination
* `command->target` is source
* Do not copy more than the destination can store


JoeNemo:
@ifakhrutdinov - please take a look.  Old code looks confusing.

ifakhrutdinov:
> @ifakhrutdinov - please take a look. Old code looks confusing.

Yeah, I meant `||`, not `&&` :-(

ifakhrutdinov (approved):
lgtm, good catch, Myt.. Martin!

PR #834, opened 2026-05-28
DatasetJson error message

Martin-Zeithaml:
1. Use `snprintf` - just for sure
2. Limit the record content to be able to print entire message:

**Record is printed completly** 
```
Record #1 with contents "RECORD WITH 81 CHARS" is longer than the max record length of 80
```
**Record is trimmned** 
```
Record #2 with contents "RECORD WITH 1025 CHARS.." is longer than the max record length of 1024
```

### DIFF
--- a/c/datasetjson.c
+++ b/c/datasetjson.c
@@ -1176,7 +1176,11 @@ static void updateDatasetWithJSONInternal(HttpResponse* response,
           if (jsonString[j] > 0x40){
             zowelog(NULL, LOG_COMP_RESTDATASET, ZOWE_LOG_DEBUG, "Invalid record for dataset, recordLength=%d but max for dataset is %d\n", recordLength, maxRecordLength);
             char errorMessage[1024];
-            int errorLength = sprintf(errorMessage,"Record #%d with contents \"%s\" is longer than the max record length of %d",i+1,jsonString,maxRecordLength);
+            if (recordLength > 900) {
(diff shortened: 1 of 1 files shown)


JoeNemo:
@Martin-Zeithaml - My original code was pretty weak.  Please add some constants.  Thanks.

JoeNemo (approved):
Approved.

c/datasetjson.c
```diff
@@ -1190,9 +1196,8 @@ static void updateDatasetWithJSONInternal(HttpResponse* response,
     }
     else {
       zowelog(NULL, LOG_COMP_RESTDATASET, ZOWE_LOG_DEBUG, "Incorrectly formatted array!\n");
-      char errorMessage[1024];
-      int errorLength = sprintf(errorMessage,"Array position %d is not a string, but must be for record updating",i);
-      errorMessage[errorLength] = '\0';
+      char errorMessage[ERROR_MESSAGE_BUFFER_SIZE];
```
ifakhrutdinov, line 1199: The content of this buffer is undetermined at this point. What happens to it, if `snprintf` returns a negative return code? Will it be still null-terminated?

Martin-Zeithaml: This change only prevents buffer overflow - which should not happen as the buffer is large and we need to format a number.

`sprintf` or `snprintf` when failing due to encoding issues, the buffer content is not defined and null terminator is not guaranteed. 

But this is completely different story and if we want to be super-safe-and-sure, we need to create new internal printing function to handle buffer overflow AND encoding issues AND ensure null terminator and include it in  every C-based Zowe repository, which is currently using `*printf*` functions family.

PR #836, opened 2026-06-04
Check `contentBody` and `contentLength`

Martin-Zeithaml:
## `c/zss.c`
* check the content body and length

## `c/safIdtService.c`
* check the content body and length
* check if returned object in not `null`

## `c/authService.c`
* check the content body and length
* check if returned object in not `null`

### DIFF
--- a/c/authService.c
+++ b/c/authService.c
@@ -200,27 +200,24 @@ static int resetPassword(HttpService *service, HttpResponse *response) {
   int returnCode = 0, reasonCode = 0;
   HttpRequest *request = response->request;
   
-  if (!strcmp(request->method, methodPOST)) {
-    char *inPtr = NULL;
-    if (request->contentBody != NULL && request->contentLength > 0) {
-      inPtr = request->contentBody;
-    }
-    char *nativeBody = NULL;
-    if (inPtr != NULL) {
-      nativeBody = copyStringToNative(request->slh, inPtr, strlen(inPtr));
-    }
-    int inLen = nativeBody == NULL ? 0 : strlen(nativeBody);
-    char errBuf[JSON_ERROR_BUFFER_SIZE];
-    char responseString[RESPONSE_MESSAGE_LENGTH];
+   if (!strcmp(request->method, methodPOST)) {
+     if (!request->contentBody || request->contentLength <= 0) {
+       respondWithJsonStatus(response, "No body found", HTTP_STATUS_BAD_REQUEST, "Bad Request");
+       return HTTP_SERVICE_FAILED;
+     }
 
-    if (nativeBody == NULL) {
-      respondWithJsonStatus(response, "No body found", HTTP_STATUS_BAD_REQUEST, "Bad Request");
-      return HTTP_SERVICE_FAILED;
-    }
-    
-    Json *body = jsonParseUnterminatedString(request->slh, nativeBody, inLen, errBuf, JSON_ERROR_BUFFER_SIZE);
-    
-    if (body == NULL) {
+     char *nativeBody = copyStringToNative(request->slh, request->contentBody,
+                                           request->contentLength);
+     if (nativeBody == NULL) {
+       respondWithJsonStatus(response, "No body found", HTTP_STATUS_BAD_REQUEST, "Bad Request");
+       return HTTP_SERVICE_FAILED;
+     }
+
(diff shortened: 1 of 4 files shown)


c/authService.c
```diff
@@ -201,8 +201,14 @@ static int resetPassword(HttpService *service, HttpResponse *response) {
   HttpRequest *request = response->request;
   
   if (!strcmp(request->method, methodPOST)) {
-    char *inPtr = request->contentBody;
-    char *nativeBody = copyStringToNative(request->slh, inPtr, strlen(inPtr));
+    char *inPtr = NULL;
```
ifakhrutdinov: I know this fix is probably driven by the initial implementation, but, well, the code is just brittle with these `inPtr` and `inLen`.

In general, the code is not ideal: why do `if (likely_scenario) { /* lots of code */ } else { /* unlikely scenario; little code */ }`, where one could just check the method at the beginning and bail if it was not POST, which would avoid this unnecessary nesting...

Would this look better perhaps?

```diff
diff --git i/c/authService.c w/c/authService.c
index 732df59..c034c06 100644
--- i/c/authService.c
+++ w/c/authService.c
@@ -201,18 +201,21 @@ static int resetPassword(HttpService *service, HttpResponse *response) {
   HttpRequest *request = response->request;

   if (!strcmp(request->method, methodPOST)) {
-    char *inPtr = request->contentBody;
-    char *nativeBody = copyStringToNative(request->slh, inPtr, strlen(inPtr));
-    int inLen = nativeBody == NULL ? 0 : strlen(nativeBody);
-    char errBuf[JSON_ERROR_BUFFER_SIZE];
-    char responseString[RESPONSE_MESSAGE_LENGTH];
+    if (!request->contentBody || request->contentLength <= 0) {
+      respondWithJsonStatus(response, "No body found", HTTP_STATUS_BAD_REQUEST, "Bad Request");
+      return HTTP_SERVICE_FAILED;
+    }

+    char *nativeBody = copyStringToNative(request->slh, request->contentBody,
+                                          request->contentLength);
     if (nativeBody == NULL) {
       respondWithJsonStatus(response, "No body found", HTTP_STATUS_BAD_REQUEST, "Bad Request");
       return HTTP_SERVICE_FAILED;
     }

-    Json *body = jsonParseUnterminatedString(request->slh, nativeBody, inLen, errBuf, JSON_ERROR_BUFFER_SIZE);
+    char errBuf[JSON_ERROR_BUFFER_SIZE];
+    char responseString[RESPONSE_MESSAGE_LENGTH];
+    Json *body = jsonParseUnterminatedString(request->slh, nativeBody, strlen(nativeBody), errBuf, JSON_ERROR_BUFFER_SIZE);

     if (body == NULL) {
       respondWithJsonStatus(response, "No body found", HTTP_STATUS_BAD_REQUEST, "Bad Request");
```

c/authService.c
```diff
@@ -200,26 +200,33 @@ static int resetPassword(HttpService *service, HttpResponse *response) {
...
-    char responseString[RESPONSE_MESSAGE_LENGTH];
+   if (!strcmp(request->method, methodPOST)) {
+     if (!request->contentBody || request->contentLength <= 0) {
+       respondWithJsonStatus(response, "No body found", HTTP_STATUS_BAD_REQUEST, "Bad Request");
+       return HTTP_SERVICE_FAILED;
+     }
 
-    if (nativeBody == NULL) {
+     char *nativeBody = copyStringToNative(request->slh, request->contentBody,
+                                           request->contentLength);
```
ifakhrutdinov, line 210: Is `contentLength` always equal `strlen(contentBody)`?

ifakhrutdinov:
@Martin-Zeithaml , could you also update `CHANGELOG.md`?

c/authService.c
ifakhrutdinov: I kind of suspect it is, but worth double checking since this would affect the logic.

ifakhrutdinov: Now I'm thinking that `contentLength` is probably the safest option because, for example, `zowe-common-c/c/httpclient.c` allocates `contentBody` in 2 places using length+1 but it never terminates that last byte, which seems like a bug (though probably latent because we never do `strlen` on `contentBody` created in `c/httpclient.c`).

Have your scans marked that source file? :-)

ifakhrutdinov (approved):

PR #837, opened 2026-06-08
Check the buffer size

Martin-Zeithaml:
Check the buffer size to prevent overflow.

### DIFF
--- a/c/zis/services/secmgmt.c
+++ b/c/zis/services/secmgmt.c
@@ -118,7 +118,13 @@ static int zisUserProfilesServiceFunctionRACF(CrossMemoryServerGlobalArea *globa
            localParmList.startUserID.length);
     const char *startUserIDNullTerm = localParmList.startUserID.length > 0 ?
                                       userIDBuffer : NULL;
-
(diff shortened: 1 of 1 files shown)


c/zis/services/secmgmt.c
```diff
@@ -118,7 +118,13 @@ static int zisUserProfilesServiceFunctionRACF(CrossMemoryServerGlobalArea *globa
            localParmList.startUserID.length);
     const char *startUserIDNullTerm = localParmList.startUserID.length > 0 ?
                                       userIDBuffer : NULL;
-
+    if (localParmList.profilesToExtract > SIZE_MAX / sizeof(RadminBasicUserPofileInfo)) {
```
ifakhrutdinov, line 121: So, have you tried allocating a size that's getting close to `SIZE_MAX`? I feel like this check will not protect against potential limits we'd reach before `SIZE_MAX` because `SIZE_MAX` is huge.

ifakhrutdinov: Though, I agree, this would protect against the integer overflow.

JoeNemo (approved):
Approved.

JoeNemo:
@Martin-Zeithaml missing changelog??

PR #841, opened 2026-06-11
Fix base64 padding length calculation

Martin-Zeithaml:
Initially there is a room for null terminator:
`allocateLength = payLoadLength + 1; // count on '\0' terminating character`

Later (for specific modulo 4 length) we are adding only space for extra equal sign(s), but not for null terminator.

```c
  allocateLength = payLoadLength + 1; // count on '\0' terminating character
  if (payLoadLength % 4 == 3) {
    allocateLength += 1;
    // space for '=' (null terminator already added)
  }
  else if (payLoadLength % 4 == 2) {
    allocateLength += 2;
    // space for '==' (null terminator already added)
  }
```


ifakhrutdinov (approved):
lgtm, thanks

PR #842, opened 2026-06-12
zss: check the path length

Martin-Zeithaml:
2 issues
1. `sprintf` can silently overwrite the buffer
2. `snprintf` is buffer-overflow safe, but we have to check, if the path was not trimmed:

If `/very/...long/path/pluginDefinition.json` would be printed as `/very/...long/path/pluginDefinition.jso` (missing `n` at the end), we will not read such file.


ifakhrutdinov (approved):
lgtm

PR #848, opened 2026-07-06
Enforce SAF authorization check for /server/agent sensitive endpoints

struga0258:
When RBAC is enabled, /server/agent/environment exposes all environment variables to any authenticated user without an actual SAF check. The code only verified that RBAC was turned on, not that the user was authorized.

Now zisCheckEntity() performs a real RACF check against ZLUX.0.COR.GET.SERVER.AGENT before granting access to sensitive data. Also added "environment" to the endpoint gate alongside config, log, and logLevels.

### DIFF
--- a/c/serverStatusService.c
+++ b/c/serverStatusService.c
@@ -42,6 +42,11 @@
 #include "configmgr.h"
 #include "serverStatusService.h"
 #include "zss.h"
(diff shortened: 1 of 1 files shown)


1000TurquoisePogs (approved):
Same as in v3

PR #850, opened 2026-07-06
Add centralized RBAC authorization for ZSS core HTTP services

struga0258:
When [dataserviceAuthentication.rbac](vscode-file://vscode-app/c:/Users/jstruga/AppData/Local/Programs/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is enabled, ZSS now performs SAF authorization checks on all core HTTP service endpoints before dispatching to the handler. The profile name follows the app-server convention: ZLUX.<instanceId>.COR.<METHOD>.<URL_SEGMENTS> in the ZOWE class. Services that don't require authentication (login, logout, saf-auth, plugins) are exempt. When RBAC is disabled, behavior is unchanged.

### DIFF
--- a/build/build_zss.sh
+++ b/build/build_zss.sh
@@ -250,6 +250,7 @@ xlc \
   ${ZSS}/c/zis/client.c \
   ${ZSS}/c/serverStatusService.c \
   ${ZSS}/c/rasService.c \
+  ${ZSS}/c/rbacService.c \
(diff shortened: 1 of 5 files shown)


1000TurquoisePogs (changes requested):
This shouldn't be touched without architectural discussion. I have started this discussion on slack.

PR #852, opened 2026-07-10
More granular security checks for built-in ZIS services

ifakhrutdinov:
## Proposed changes

The purpose of the new checks is more granular security since some users may not want to give access to all the built-in services by default; this is similar to the mechanism that already exists in ZIS plug-ins.

The implementation is not generic as it just adds explicit checks to each service. This was done on purpose as it is very unlikely we'd be adding more core services to ZIS as the plug-in system is more flexible (similar how you'd choose PCs over SVCs for your application). A generic implementation has been attempted, but it adds unnecessary code to the
common path and has to change the xmem global area which all may affect users who don't care about the built-in services.

All the services except AUTH have the new SAF checks enabled by default. The user can specify a special per-service PARMLIB parameter to disable the corresponding SAF check in case granting additional SAF access is not possible.

Below is the table listing the SAF profiles in the FACILITY class and options for each service:
| Service               | SAF profile                   | Required access | PARMLIB option              | Default value |
|-----------------------|-------------------------------|-----------------|-----------------------------|---------------|
| Auth                  | ZWES.IS.SRV.AUTH              | UPDATE          | ZWES.SRV.AUTH.SAF           | OFF          |
| NWM                   | ZWES.IS.SRV.NWM               | READ            | ZWES.SRV.NWM.SAF            | ON            |
| Snarfer               | ZWES.IS.SRV.SNARFER           | READ            | ZWES.SRV.SNARFER.SAF        | ON            |
| User profiles         | ZWES.IS.SRV.SM.USRPROF        | READ            | ZWES.SRV.SM.USRPROF.SAF     | ON            |
| Genres profiles       | ZWES.IS.SRV.SM.GENPROF        | READ            | ZWES.SRV.SM.GENPROF.SAF     | ON            |
| Genres access list    | ZWES.IS.SRV.SM.GENACL         | READ            | ZWES.SRV.SM.GENACL.SAF      | ON            |
| Genres profile admin  | ZWES.IS.SRV.SM.GENADM         | UPDATE          | ZWES.SRV.SM.GENADM.SAF      | ON            |
| Group profiles        | ZWES.IS.SRV.SM.GRPPROF        | READ            | ZWES.SRV.SM.GRPPROF.SAF     | ON            |
| Group access list     | ZWES.IS.SRV.SM.GRPACL         | READ            | ZWES.SRV.SM.GRPACL.SAF      | ON            |
| Group admin           | ZWES.IS.SRV.SM.GRPADM         | UPDATE          | ZWES.SRV.SM.GRPADM.SAF      | ON            |

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## PR Checklist
- [x] I have added tests that prove my fix is effective or that my feature works, or describe a test method below

## Testing

Use the new test program (`tests/zis-saf-tests.c`) to verify that the additional SAF checks and PARMLIB config parameters work as intended.

### DIFF
--- a/h/zis/services/common.h
+++ b/h/zis/services/common.h
@@ -15,9 +15,17 @@
 
 #define ZIS_SERVICES_DEFAULT_SAF_CLASS "FACILITY"
 
+/**
+ * The purpose of this structure is to communicate additional flags to the core
+ * services.
+ *
+ * The struct is passed to the core services within the storage of the "parm"
+ * 8-byte pointer passed to service functions; this is why it must always fit in
+ * 8-bytes. The reason for this hack is to avoid allocating additional common storage.
+ */
 typedef struct ZISCoreServiceParm_tag {
-#define ZIS_CORE_SERVICE_FLAG_NO_SAF_CHECK 0x00000001
-  unsigned flags : 8;
+#define ZIS_CORE_SERVICE_FLAG_NO_SAF_CHECK 0x01
+  unsigned char flags;
   char reserved[7];
 } ZISCoreServiceParm;
 
--- a/h/zis/services/common.h
+++ b/h/zis/services/common.h
@@ -19,9 +19,10 @@
  * The purpose of this structure is to communicate additional flags to the core
  * services.
  *
- * The struct is passed to the core services within the storage of the "parm"
- * 8-byte pointer passed to service functions; this is why it must always fit in
- * 8-bytes. The reason for this hack is to avoid allocating additional common storage.
+ * The struct is passed to the core services within the storage of the
+ * "CrossMemoryService->serviceData" 8-byte pointer passed to service
+ * functions; this is why it must always fit in 8-bytes. The reason for this
+ * hack is to avoid allocating additional common storage.
  */
 typedef struct ZISCoreServiceParm_tag {
 #define ZIS_CORE_SERVICE_FLAG_NO_SAF_CHECK 0x01
--- a/h/zis/services/common.h
+++ b/h/zis/services/common.h
@@ -1,0 +1,39 @@
+
+
+/*
+  This program and the accompanying materials are
+  made available under the terms of the Eclipse Public License v2.0 which accompanies
+  this distribution, and is available at https://www.eclipse.org/legal/epl-v20.html
+
+  SPDX-License-Identifier: EPL-2.0
+
+  Copyright Contributors to the Zowe Project.
+*/
+
+#ifndef ZIS_SERVICES_COMMON_H_
+#define ZIS_SERVICES_COMMON_H_
+
+#define ZIS_SERVICES_DEFAULT_SAF_CLASS "FACILITY"
+
+typedef struct ZISCoreServiceParm_tag {
+#define ZIS_CORE_SERVICE_FLAG_NO_SAF_CHECK 0x00000001
+  unsigned flags : 8;
+  char reserved[7];
+} ZISCoreServiceParm;
+
+#define IS_ZIS_CORE_SERVICE_SAF_ON(data) ({ \
+     !(((ZISCoreServiceParm *)&data)->flags & ZIS_CORE_SERVICE_FLAG_NO_SAF_CHECK); \
+   })
+
+#endif /* ZIS_SERVICES_COMMON_H_ */
+
+
+/*
+  This program and the accompanying materials are
+  made available under the terms of the Eclipse Public License v2.0 which accompanies
+  this distribution, and is available at https://www.eclipse.org/legal/epl-v20.html
+
+  SPDX-License-Identifier: EPL-2.0
+
+  Copyright Contributors to the Zowe Project.
+*/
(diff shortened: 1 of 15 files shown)


h/zis/services/common.h
```diff
@@ -0,0 +1,39 @@
...
+  Copyright Contributors to the Zowe Project.
+*/
+
+#ifndef ZIS_SERVICES_COMMON_H_
+#define ZIS_SERVICES_COMMON_H_
+
+#define ZIS_SERVICES_DEFAULT_SAF_CLASS "FACILITY"
+
+typedef struct ZISCoreServiceParm_tag {
+#define ZIS_CORE_SERVICE_FLAG_NO_SAF_CHECK 0x00000001
```
Martin-Zeithaml: Should we add "pragma pack" here?
Will this define 4 (unsigned) + 7 (char[7]) + 1 (padding)
or
1 (unsigned)  + 7 (char[7]) bytes?

ifakhrutdinov: The listings tell that it's all 8-bytes, but let me double check with the doc...

ifakhrutdinov: From "z/OS XL C/C++ Language Reference":
> Bit fields are bit packed. They can cross word and byte boundaries. No padding is
inserted between two (non-zero length) bit field members. Bit padding can occur
after a bit field member if the next member is a zero length bitfield or a non-bit
field. Non-bit field members are aligned based on their declared type. 

Since the bit field, which is 1 byte, is followed by a char array, there is no padding.

Martin-Zeithaml: This code is fine and will wok, but it is not resistant for future updates.

The `char[7] reserved` is actually used for padding (at least I think) and should not be used by anyone in the future.
Can we rename it or add some comment explaining the padding?

ifakhrutdinov: > The char[7] reserved is actually used for padding (at least I think) and should not be used by anyone in the future.

Actually, in this case, `reserved` is both to indicate that this storage can be used for something in the future and that it makes the struct match the storage that backs it.

A real problem I see is that the `define` for the only flag is actually 4 bytes, whereas the flag field is 1 byte. I'll change that.

> This code is fine and will wok, but it is not resistant for future updates.

Could you give an example of an update that'd break this? BTW, I really hope we don't need to update these services in the future; we should eventually move them out of the core to some optional plug-ins.

Martin-Zeithaml:
```c
#include <stdio.h>

#define ZIS_CORE_SERVICE_FLAG_NO_SAF_CHECK 0x01

typedef struct ZISCoreServiceParm_tag {
  unsigned flags : 8;
  char reserved[7];
} ZISCoreServiceParm;

void *testThis(int addReserved) {
  union {
    ZISCoreServiceParm aStr;
    void *asPtr;
  } parm = {0};
  
  parm.aStr.flags |= ZIS_CORE_SERVICE_FLAG_NO_SAF_CHECK;
  if (addReserved) {
    parm.aStr.reserved[0] = ' ';
  }
 
  return parm.asPtr;
}

int main() {
    
    printf("%p\n", testThis(0));
    printf("%p\n", testThis(1));

    return 0;
}
```

Output (compiled and executed outside mf):
```
0x1
0x2001
```

ifakhrutdinov: Isn't this expected? You've started using that reserved field, so the results are different.

This struct and services are tightly coupled and when someone decides to use the reserved field, the services would need to be changed accordingly.

Or do I not get this example?

ifakhrutdinov: Or is it about getting `0x2001` instead of `0x0120`? I think it's just how your machine assume endianness: `%p` in `printf` will produce a hex string with most significant digit on the left (regardless of the architecture) because, according to [this](https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Printing-Pointers.html), it'll treat it as an unsigned int of an appropriate size; on your machine, which is probably little-endian, the most significant digit is in `reserved[6]` instead of `flags`, so `printf` prints the reverse of what you expect: `0x0000000000002001` or `0x2001` when the leading zeroes are truncated.

In memory it's still in the expected order though; you can add these lines to check that:

```c
    void *data = testThis(1);
    printf("flags=0x%02x\n", ((ZISCoreServiceParm *)&data)->flags);

```

Martin-Zeithaml: I was using AI to review this pull request and wanted to double-check the results. I figured it is better to ask and ensure we avoid any potential bugs in the code.

Thanks for the explanation!

Martin-Zeithaml (dismissed):

ifakhrutdinov:
@Martin-Zeithaml , FYI, see https://github.com/zowe/zss/compare/168fa3974f28a0dc28e1eb588c42327f0a426405..839beac2a9fdaea1f7be452d8c90b8c88ab1562c

I added a comment and changed that bit field to be `unsigned char` since the latter should be less confusing. I also added forward declarations of the parm struct because for some reason I didn't notice the respective info messages from the compiler.

Please have a look.

JoeNemo (dismissed):
Approved.

JoeNemo (approved):

PR #866, opened 2026-08-31
Print the `3.x` for specific error

Martin-Zeithaml:
In V3, when `zss` is started without schema or config, it prints the message `ZWES1013I` with the version.
Following message is misleading.

```
ZWES1013I ZSS Server has started. Version '3.5.0+20260501' 64-bit
ZSS 2.x requires schemas and config
```

Alternative (future version independent) solution would be `ZSS 2.x or later requires schemas and config`


JoeNemo (approved):
Approved.

PR #873, opened 2026-09-08
Support verifyCertificates modes: STRICT and NONSTRICT

ChongZhou-Broadcom:
Support configuration `zowe.verifyCertificates`.

**Currently, only `STRICT` and `NONSTRICT` are supported.**

ZSS only reads the related configuration and set TLS env. `zowe-common-c` does the real work.

`DISABLED` is not supported for now, but ZSS only propagates the configuration, `zowe-common-c` will force it to `NONSTRICT` and write `LOG_COMP_HTTPCLIENT` log.

See related PR in zowe-common-c: [PR#616 Let GSK validate the CN and SAN domains of server certificates- #616
](https://github.com/zowe/zowe-common-c/pull/616/changes)


1000TurquoisePogs (commented):
This looks perfect, awaiting testing on zowe-common-c part.

PR #878, opened 2026-09-16
[V2] Use proper variable for `findSessions`

Martin-Zeithaml:
Looks like copy-paste bug:

```c
  if (luname = getQueryParam(request,"luname")){                  // If luname = NULL
    findSessions(context,SESSION_KEY_TYPE_LUNAME,0,luname);
  } else if (ipAddress = getQueryParam(request,"ip4Address")){    // And ipAddress=something
    findSessions(context,SESSION_KEY_TYPE_IP4_STRING,0,luname);   // We could use NULL value here
  } else if (ipAddress = getQueryParam(request,"ip6Address")){
    findSessions(context,SESSION_KEY_TYPE_IP6_STRING,0,luname);   // Or here
  } else{
    findSessions(context,SESSION_KEY_TYPE_ALL,0,luname);
    zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_DEBUG, "get'em all\n");
  }
```
* Use `ipAddress` instead of possibly NULL `luname`
* Else part updated for readability (not a bug, because SESSION_KEY_TYPE_ALL ignores the string key)


JoeNemo (approved):
Approved.

PR #881, opened 2026-09-17
[V2] NULL check after `getQueryParam`

Martin-Zeithaml:
Note: Skip checking of `pattern` in `serveUnixFileChangeMode`, this parameter could be NULL (noted in comment).


JoeNemo (approved):
Approved.

PR #882, opened 2026-09-17
[V3] NULL check after `getQueryParam`

Martin-Zeithaml:
* Note: Skip checking of `pattern` in `serveUnixFileChangeMode`, this parameter could be NULL (noted in comment).
* Identical to #881


JoeNemo (approved):
Approved.

